Issue #62, opened 2016-01-08
Move from cron to new scheduled tasks system


micaherne:
It would be good if this plugin could support the new scheduled tasks system that was introduced in Moodle 2.7 (https://docs.moodle.org/30/en/Scheduled_tasks), which supersedes the legacy cron system.

micaherne:
Sorry, I've just discovered that there is a core Moodle task \core\task\plagiarism_cron_task that calls the cron() method for each plugin so this isn't correct.

danmarsden:
yeah - I've seen the same issue as I've switched the URKUND plugin over to use scheduled tasks too - I see your patch to fix in core, I'll review and get that pushed through later this week! - thanks.

sk-unikent:
I believe it does still use the Event 1 system which does, in turn, use the old queuing system.
(see db/events.php)

danmarsden:
@sk-unikent you might want to report a separate issue for that (if one doesn't already exist) - I've just migrated all the URKUND code from the legacy events structure to the new events 2 api too.

PR #63, opened 2016-01-08
Use scheduled tasks system

micaherne:
Hi John, here's a patch for #62 (against develop branch). It's not particularly well-tested but should be most of the way there. I haven't been able to test that the legacy cron job only runs on versions < 2.7, but I think the logic there is sound. Hope it's useful!

Cheers, Michael

### DIFF
--- a/classes/task/update_scores.php
+++ b/classes/task/update_scores.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace plagiarism_turnitin\task;
+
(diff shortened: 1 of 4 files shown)


micaherne:
Sorry, I've just discovered that there is a core Moodle task \core\task\plagiarism_cron_task that calls the cron() method for each plugin, so this patch isn't necessary.

Issue #115, opened 2016-03-17
Submissions "Pending"


mhughes2k:
We're seeing some cases where mod_assign submissions are displaying a "Pending" status for the plagiarism plugin but they don't appear to ever be coming back.

We think that the academic member of staff enabled Turnitin *after* students had started submitting items.

Would items that have already been submitted, be sent to Turnitin if it is subsequently turned on. Looking through the code I can't find a mechanism that adds them to the `turnitin_files` table unless they are actually edited and raises one of `file_uploaded`, `assessable_submitted`, `content_upload` or `files_done` events. 

The `mod_created` and `mod_updated` events do not appear to check existing for submissions.

jmcgettrick:
Hi @mhughes2k, no they wouldn't be sent in that instance, they're only sent from events.
The "pending" is - at best - confusing in that case as those are never going to be sent. 

Ideally we'd be able to work out whether there is an event for that submission that hasn't yet been processed. However, when I've looked in the past I haven't been able to work out an easy way of doing that.

mhughes2k:
The problem appears to be not that the submission hasn't been processed; it has, but it was processed *before* TII was turned on.

Given there is a "Refresh Submissions" button couldn't that simply scan Moodle assignment submissions and check if there is a corresponding turnitin_files record, and if there isn't it would indicate that the submission needs to be sent?

I've noted that Moodle appears to raise the mod_updated and mod_Created events *before* plagiarism plugins have been passed the updated / created module information, so hooking the above into these events I would presume won't work because the updated information won't have reached the plugin (see end of `edit_module_post_actions()` function in modlib.php)

mhughes2k:
The other thing that I've discovered is that I have gone in to the assign activity, reverted the submission to being a draft and re-submitted it, which has changed the timestamp, and it is logged in Moodle that I've updated the submission (so the event should have been raised)

But despite this the item is still sitting as pending, and isn't listed in the `turnititin_files` table.

mhughes2k:
Also I've asked the question about the `mod_updated` event being raised before the plagiarism plugin has been able to save the data at https://moodle.org/mod/forum/discuss.php?d=330029

jmcgettrick:
Is the resubmitted submission not raising an event in the events queue at all?

mhughes2k:
So, yesterday I reverted the submission to draft and then re-submitted it.

[...]

PR #117, opened 2016-03-18
Added check for 0 length turnitin_files to see if never submitted

mhughes2k:
I think that something like this would address the clarity of #115


davidmhatton:
Hi @mhughes2k I have created a slight modification of this pull request on our develop branch that is in line with our conventions etc.
Thanks for your contribution!

PR #136, opened 2016-05-09
Fixes #135 Default settings not applied if setting not locked.

### LINKED ISSUE #135: Default settings not applied if not "locked"
The default settings in plagiarism/turnitin/settings.php?do=defaults only appear to be used if the setting is also set to "locked" when creating new assignments.

it looks like the function get_settings() is used to return the defaults but this function only returns values if the field is locked.

danmarsden:


jmcgettrick:
Hi @danmarsden - I've applied this to develop branch in https://github.com/turnitin/moodle-plagiarism_turnitin/pull/139.

Thanks.

PR #137, opened 2016-05-09
Change recordset back to records

jmcgettrick:
This is causing problems with the display of the data dump for a lot of users.


davidmhatton:
👍

PR #142, opened 2016-05-16
Check for $CFG->version before using

micaherne:
The test for the Moodle version number in version.php causes a PHP notice to be thrown when $CFG->version is not set (e.g. when running unit tests):

> Notice: Undefined property: stdClass::$version in F:\htdocs\pixie-beta\plagiarism\turnitin\version.php on line 15

 It would be good to test for this property's existence before using it.

### DIFF
--- a/version.php
+++ b/version.php
@@ -12,7 +12,7 @@ $plugin->component = 'plagiarism_turnitin';
 $plugin->maturity  = MATURITY_STABLE;
 
 global $CFG;
-if ($CFG->version > 2014051200) {
+if (!empty($CFG->version) && $CFG->version > 2014051200) {
 	$plugin->cron = 0;
 }else{
 	$plugin->cron = 300;


danmarsden:
yeah - this is a big issue for us too. I'd recommend Turnitin add travis-ci integration to their plugins to catch this sort of thing during release if possible:
David has a good post on this here:
https://moodle.org/mod/forum/discuss.php?d=323384

jmcgettrick:
Thanks @danmarsden - That looks good, we'll take a look.

danmarsden:
The first time I added it to a plugin I didn't quite believe how simple it was.... you just grab the template
https://github.com/moodlerooms/moodle-plugin-ci/blob/master/.travis.dist.yml
modify it to suit php versions/moodle versions etc and the reports it spits out are great. Really nice to have the php7 stuff and behat all run automatically on all pull requests in github too!

jmcgettrick:
This has been added in as of 2016072601. Thanks for the PR but it was to master and w had to make the changes to develop for QA first.

Issue #89, opened 2016-02-18
EULA opens in same window- UX problem


mkpelletier:
248fbf1 seems to have created a UX problem, whereby a new user is directed in the same window as an assignment submission in the LMS.  After accepting the EULA, the user finds himself at a dead end.  The page where he started is in the history barricaded by a redirect on Turnitin's own website.  Click back just brings him back to the EULA, unless he frantically clicks back several times.  Didn't this used to all happen in a lightbox like the Rubric.  Why has this been changed to open in the same window?

jmcgettrick:
Is this still happening @mkpelletier aswe can't reproduce it?

mkpelletier:
Hello John,
This is still happening.  Please refer here to see it in action:http://www.youtube.com/watch?v=SGqXf4URXzA

jmcgettrick:
I can't reproduce this do you get a js error in firebug? I think there's maybe a conflict.

pauln:
We encountered this while testing the latest version of the plugins before rolling them out to production.  @jmcgettrick, you might have been having difficulty reproducing it because (as per the video posted by @mkpelletier) it doesn't seem to be a problem if the student accepts the EULA when first presented, but if they initially decline the EULA, upload a file to the assignment, _then_ click on the "click here to accept" link either from the submission status page or the edit submission page, it (almost always) triggers this issue.  No errors appear in the Firebug / dev tools console, which makes it a bit tricky to diagnose.

The problem is that colorbox loads the content returned from extras.php inline, instead of in an iframe; I haven't managed to determine why this happens, and it doesn't appear to have been fixed in a more recent version of colorbox.  Because that content is a form (with a JS call to submit said form), one of two things happens when it's displayed inline:
- As described here, the EULA form replaces the Moodle site, leaving the user "stuck"; or
- A different form gets submitted, because it's submitting the first form on the page (`window.document.forms[0]`); depending on the form, this can lead to a TII page opening in a new tab/window and prompting for a username/password via HTTP authentication

Although I couldn't figure out why it was happening, I did manage to find a workaround: rather than calling colorbox on the link directly, add a click handler which then triggers colorbox.  I've submitted a pull request with this fix (#148).  It may be that there's some kind of race condition happening, as various combinations of connection throttling and breakpoints seemed to "fix" it sometimes.

dwinn:
Hi @pauln,

[...]

PR #148, opened 2016-06-01
Wrap colorbox in a click handler for EULA link

pauln:
When calling colorbox directly on the EULA link, it doesn't always open
in an iframe if the student has previously declined the EULA - often it
will load the form inline, resulting in the browser being redirected to
the EULA on a blank page on the TII site instead of displaying on top of
the Moodle page.  This means that the user gets "stuck" on the TII site -
they're not returned to Moodle after either accepting or declining.

This fixes issue #89.


dwinn:
Hi @pauln Thanks for this pull request. It works in my initial check so I'll put it through to our QA and we'll get this out in the next release.

PR #150, opened 2016-06-28
Add check for file contents

kenneth-hendricks:
Had a system outage and the event handler was trying to fetch the contents of a non existent file. Uncaught exception halted any submissions from being processed. This adds a check that functions similarly to the existing file check above it.


jmcgettrick:
Hi @kenneth-hendricks  - could you resubmit this pull request please, to the develop branch and we'll get it merged in. Thanks!

PR #152, opened 2016-06-28
Closes #151: Fixed bug where users saw default error messages instead of proper ones

### LINKED ISSUE #151: Bug in error messages
We noticed that lecturers were getting the standard "This file has not been submitted to Turnitin, please consult your system administrator" error and students were seeing "This file has not been submitted to Turnitin, please consult your tutor for further details" error where in fact going to the error tab in TII settings showed the actual reason for the error (either there weren't enough words, or a wrong format).

marcusboon:


donkeyx:
+10 for this as those error messages are not very useful

PR #162, opened 2016-07-25
INT-6179 get records fix

davidmhatton:
- change data query to use `get_records()` instead of `get_recordset()`


jmcgettrick:
Closing. Already been done in 2016072601.

Issue #179, opened 2016-10-13
PHPUnit test failing with Moodle 31


roperto:
Has anyone else run into this issue?

When I run core Moodle 31 PHPUnit tests, one of them fails if I have this plugin installed.

I've isolated the problem in the lib, where it finds the supported modules based on the file (classes/modules) system then tries to fetch the ID in the database. 

Because it cannot find the module in the DB (returns false), if fails when trying go get the $module->id.

I wrote a test reproducing the behaviour and a proposed solution; however, someone who has better understanding of this plugin could propose a better solution.

Cheers,

Daniel

danmarsden:
This is a pretty major issue for us... we run these tests automatically on all our clients codebases.

 It might be worthwhile for the Turnitin team to implement travis-ci checks directly within github to help prevent this stuff from happening.
There's a good post on how to add support here:
https://moodle.org/mod/forum/discuss.php?d=323384

but it's really just copying a .travis.yml file, adjusting it as needed and going to travis-ci.org and switching it on.

alexumn:
A quick note: https://github.com/turnitin/moodle-plagiarism_turnitin/pull/188/ (created by one of our Moodle developers, Travis Noll) addresses this specific issue as well. Let this be another vote to implement travis-ci and Behat tests in this plugin.

micaherne:
We're also seeing this on Moodle 3.0:

1) tool_uploadcourse_course_testcase::test_reset
Unexpected debugging() call detected.
Debugging: Exception encountered in event observer 'plagiarism_plugin_turnitin::course_reset': Trying to get property of non-object
* line 328 of /plagiarism/turnitin/lib.php: call to PHPUnit_Util_ErrorHandler::handleError()
* line ? of unknownfile: call to plagiarism_plugin_turnitin::course_reset()
* line 155 of /lib/classes/event/manager.php: call to call_user_func()
* line 75 of /lib/classes/event/manager.php: call to core\event\manager::process_buffers()
* line 804 of /lib/classes/event/base.php: call to core\event\manager::dispatch()
* line 5306 of /lib/moodlelib.php: call to core\event\base->trigger()
* line 902 of /admin/tool/uploadcourse/classes/course.php: call to reset_course_userdata()
* line 748 of /admin/tool/uploadcourse/classes/course.php: call to tool_uploadcourse_course->reset()
[...]

PR #180, opened 2016-10-13
Issue #179 - PHPUnit test failing with Moodle 31

roperto:
Has anyone else run into this issue?

When I run core Moodle 31 PHPUnit tests, one of them fails if I have this plugin installed.

I've isolated the problem in the lib, where it finds the supported modules based on the file (classes/modules) system then tries to fetch the ID in the database.

Because it cannot find the module in the DB (returns false), if fails when trying go get the $module->id.

I wrote a test reproducing the behaviour and a proposed solution; however, someone who has better understanding of this plugin could propose a better solution.

Cheers,

Daniel


roperto:
Created new pull request from develop.

PR #186, opened 2016-11-10
plagiarism_turnitin: Migrate Event 1 API to Event 2

tlock:
- Updated events to use the new API
 - Remove all associated legacy events

### DIFF
--- a/db/events.php
+++ b/db/events.php
@@ -14,48 +14,43 @@
 // You should have received a copy of the GNU General Public License
 // along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
 
-$handlers = array(
-    'assessable_file_uploaded' => array(
-        'handlerfile'       => '/plagiarism/turnitin/lib.php',
-        'handlerfunction'   => 'plagiarism_turnitin_event_file_uploaded',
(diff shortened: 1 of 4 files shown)


roperto:
:+1: Hi Tim, thanks for the migration, I am cherry-picking this until it arrives in master :-)

roperto:
Just a late feedback, but some changes in the lib.php had a few problems here. Some of the event data parameters differ from the original one, I did not investigate further but it could be something related to:

https://github.com/turnitin/moodle-plagiarism_turnitin/pull/186/files#diff-828043dc9f8d10ccd58ff8b96297794fR2050

I think turnitin has several changes in their develop branch including database migrations, so for now we are still using event 1 API and waiting until they release it stable.

Cheers,
Daniel

tlock:
Unfortunately, this hasn't solved all the issues so we'll leave it till we see what Tii do with the deprecated issues.

amclarty:
@tlock, @roperto We're working on the Event 2 changes now. Thanks for your input.

dwinn:
Hi @tlock ,

We have just released v2015012501, which contains our Event 2 API changes. Thanks for your input on the issue.

PR #187, opened 2016-11-14
Moodle Behat tests fail when a module isn't installed 

travisnoll:
Moodle Behat tests fail when a module isn't installed but lib.php calls $module->id

Error message from behat:

PHP debug message/s found:
Notice: Trying to get property of non-object in /home/oit-moodle/3.0/docs/plagiarism/turnitin/lib.php on line 328

Failing Moodle Behat feature files:

mod/chat/tests/behat/chat_course_reset.feature
mod/lesson/tests/behat/lesson_course_reset.feature
mod/quiz/tests/behat/quiz_reset.feature

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -318,6 +318,11 @@ class plagiarism_plugin_turnitin extends plagiarism_plugin {
         foreach ($supportedmods as $supportedmod) {
             $module = $DB->get_record('modules', array('name' => $supportedmod));
 
+            if (!$module){
+               #Skip module that isn't installed.
+               continue;
+            }
+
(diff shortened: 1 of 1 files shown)


danmarsden:
I'd close this one as a duplicate of #183 

Also FYI - looks like the patch doesn't meet Moodle coding guidelines - comments should use // not #, spacing around if () { not correct, and a patch should only change the necessary lines - it's changing the line ending in the file as well.

jmcgettrick:
Thanks @danmarsden, #183 has been merged.

Issue #194, opened 2016-12-01
Turnitin slows mod/forum/post.php even when TII is not enabled for that mod.


roperto:
A call to mod/forum/post.php may trigger a print_disclosure() for all plagiarism plugins.

If the TII server is slow (or there is a timeout), the page becomes very slow even when turn it in is not needed.

roperto:
Hi,

I created pull requests to both master and develop as I am not sure how workflow with TII.

Cheers,
Daniel

danmarsden:
afaik they go to develop first and then TII pushes through to master on release.

roperto:
Thanks dan, when (if) they merge one then I will remove the other request :-)

jmcgettrick:
Fixed in v2017052401

PR #195, opened 2016-12-01
(master) Fixes Issue #194 - Turnitin slows mod/forum/post.php even when TII is not enabled for that mod.

roperto:
A call to mod/forum/post.php may trigger a print_disclosure() for all plagiarism plugins.

If the TII server is slow (or there is a timeout), the page becomes very slow even when turn it in is not needed.

This fix: print_disclosure() now only checks the connection when it is needed.


roperto:
This pull request is wrong (to master), leaving only the other one (to develop).

Issue #217, opened 2017-02-05
Undefined property: turnitintooltwo_user::$useragreementaccepted


roperto:
`Notice: Undefined property: turnitintooltwo_user::$useragreementaccepted in .../plagiarism/turnitin/lib.php on line 431`

I am getting this error when submitting an assignment.

jmcgettrick:
Hi @roperto - is this in Moodle 3.2?

roperto:
Hi @jmcgettrick 

I am using it on Moodle 3.1,

Cheers,

Daniel

dwinn:
Hi @roperto,

We changed some variable names in the last release as a result of Travis linting suggestions. PP depends on Moodle Direct V2, did you also update this plugin? This could explain the variable names not matching up.

roperto:
Hi. I didn't update Moodle Direct V2 (what is it?) but I did forget to update `mod/turnitintooltwo`

After updating `mod/turnitintooltwo` it worked, so I am closing this issue.

Thanks for the feedback,

Daniel

dwinn:
Hi @roperto,

Glad to hear you have resolved the issue.

Moodle Direct V2 is a separate plugin which creates a Turnitin assignment module, whereas the Plagiarism Plugin adds Turnitin functionality to the Moodle assignment module. As they both connect to Turnitin they share a fair bit of code.

PR #218, opened 2017-02-05
Fixes issue #217 - useragreementaccepted is undefined..

roperto:
The correct name of the property is: user_agreement_accepted


roperto:
The issue was fixed by update turnitintooltwo. This pull request is not needed.

PR #225, opened 2017-02-07
Fixes #179 Behat test failure.

### LINKED ISSUE #179: PHPUnit test failing with Moodle 31

danmarsden:
I think this should fix it - I don't like the way this function looks all that much but this retains the existing logic. (just re-running all the behat tests on our systems - will report back if it's still failing.)

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -301,11 +301,18 @@ class plagiarism_plugin_turnitin extends plagiarism_plugin {
         $courseid = (int)$data['other']['reset_options']['courseid'];
         $resetcourse = true;
 
-        $resetassignsubmissions = $data['other']['reset_options']['reset_assign_submissions'];
-        $resetassign = (isset($resetassignsubmissions)) ? $resetassignsubmissions : 0;
-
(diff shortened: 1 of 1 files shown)


danmarsden:
forgot to mention - all behat tests pass with this patch in place.

dwinn:
Hi @danmarsden ,

Thanks for the pull request, it looks like it should work but we will have to put it through our QA process.

From the looks of it we broke this when we implemented Travis. Travis was complaining that the line was too long so we set the variable $data['other']['reset_options']['reset_assign_submissions'] to $resetassignsubmissions first. This threw off the isset() check. Likewise with the second one.

PR #252, opened 2017-06-08
Ignore missing course modules

micaherne:
Hi, here's a patch to add a bit of defensive programming to prevent missing course modules crashing the send_submissions task. I know the plugin should remove files when a course module is deleted, but we've seen a couple of instances where there are still files for a deleted module (not sure why). In line with the rest of Moodle it's a good idea not to assume that a cm exists, even when it "should".


jmcgettrick:
Cheers @micaherne - we've extended this slightly, should be released soon: https://github.com/turnitin/moodle-plagiarism_turnitin/pull/254

Issue #250, opened 2017-05-31
EULA issue with already accepted users 


hitteshahuja:
Using version 2016011105 . 
Installed the plugin on a brand new Moodle site. Setup correctly as Moodle can connect to tii without any issues. When trying to submit the file as an existing user, I always keep getting 

**This file has not been submitted to Turnitin because the user has not accepted the Turnitin End User Licence Agreement.**
This is incorrect because the user I am testing has already accepted the eula for the turinitin service, albeit on a different moodle site. Moreover, I dont see a link presented by the plugin to accept the EULA again. 

the api logs seems fine for the same  . I see classes created and users sent through but their submissions are not being accepted it seems. 

```
2017-05-31 15:49:06 +0100 (REQUEST) - Part created as Turnitin Assignment (789xxxx) - Tool Id: (0) - Part num: (0)
2017-05-31 15:49:24 +0100 (REQUEST) - User 3 (400xxxx) joined to class (3496xxx)
2017-05-31 15:50:08 +0100 (REQUEST) - Turnitin Assignment part updated - id: 789xxxx
2017-05-31 15:50:12 +0100 (REQUEST) - Turnitin Assignment part updated - id: 789xxxx
2017-05-31 15:59:05 +0100 (REQUEST) - Turnitin Assignment part updated - id: 789xxxx
2017-05-31 15:59:08 +0100 (REQUEST) - Turnitin Assignment part updated - id: 789xxxx
2017-05-31 16:02:09 +0100 (REQUEST) - Turnitin Assignment part updated - id: 789xxxx
2017-05-31 16:02:13 +0100 (REQUEST) - Turnitin Assignment part updated - id: 789xxxx
2017-05-31 16:02:59 +0100 (REQUEST) - User 4 (5515290) joined to class (3496xxx)
```

hitteshahuja:
![image](https://cloud.githubusercontent.com/assets/2855029/26640459/af5b41ce-461f-11e7-87c3-4bcc7fcd9b20.png)

jmcgettrick:
Hi, we can't reproduce this here. We've released six updated plugins since 2016011105 though, can you reproduce on the current version?

hitteshahuja:
Hi John,
Yes, we can reproduce it on 2017052401. This is what we have done to reproduce the error :

[...]

PR #255, opened 2017-06-21
Potential fix for #250

hitteshahuja:
Hi John,
I believe this is what should do the trick. We are making sure that if a student ignores the EULA the first time around but accepts it retrospectively (for a separate assignment for example) then their previous submissions should be queued again and sent to turnitin. 

PS  - I'm sure there is a better way to get the cmid to pass to ajax.php when a user accepts the EULA by clicking on 'I agree' -  [code](https://github.com/hitteshahuja/moodle-plagiarism_turnitin/blob/addff721330a85d665d3a573f455f8c4e3e9b0a3/jquery/turnitin_module.js#L135 )

### DIFF
--- a/ajax.php
+++ b/ajax.php
@@ -207,6 +207,13 @@ switch ($action) {
 
         // Update the user using the above object.
         $DB->update_record('turnitintooltwo_users', $eulauser, $bulk = false);
+
+        //Also reset their submissions ( ? ) University of Bath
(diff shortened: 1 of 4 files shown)


cwestgarth:
Hi Hittesh,

Thanks for the pull request, however as John mentioned this is not something which we will be able to implement. We had to create this workflow to align with a privacy legislation. We cannot assume that the EULA was ignored and the student did not have a legitimate reason to decline the EULA in the past.

Thanks again

Caroline Westgarth
Product Manager, Integrations

Issue #264, opened 2017-08-07
Submission cron attempts to use a private method and fails


aolley:
It looks like when the plugin moved to implementing a queue to submit files, that the 
`delete_tii_submission` method of the `plagiarism_plugin_turnitin` class was made private.

However, there were still calls to that function from outside the class which weren't changed to account for this. So once there's a submission that trips that particular code path, the send_submissions cron task becomes permanently stuck.

This is possibly the cause of #240.

dwinn:
Hi @aolley,

Thank you for reporting this issue. I see what you mean, `plagiarism_turnitin_send_queued_submissions` is calling `delete_tii_submission` from outside the `plagiarism_plugin_turnitin` class. We will get it looked at as it sounds like you may have found a bug.

Also thank you for the pull request, but would you be able to make the pull request to our develop branch instead of master? We do our testing on develop branch before merging to master for releases.

aolley:
Hi @dwinn - sorry about that. PR'd against the correct branch.

dwinn:
Hi @aolley,

Thank you for the new pull request. I've tested the change and it works. We are planning a small release this week so we'll throw it in.

dwinn:
Hi @aolley,

We have just released v2017081001 with this fix.

PR #266, opened 2017-08-07
Keep delete_tii_submission public till its no longer called externally

aolley:
This function is used from outside the class its in during cron, so we
can't have it private till thats fixed up.

Without this change, a submission triggering that call can cause Moodle
cron to get blocked.


aolley:
This is for issue #264

PR #273, opened 2017-10-12
Fix to check the plugin is configured

jmcgettrick:


tests/lib_test.php
```diff
@@ -0,0 +1,61 @@
...
+        $this->assertEquals(false, $ispluginconfigured);
+
+        // Check if plugin is configured with only account id set.
+        set_config('accountid', '1001', 'turnitintooltwo');
+        $ispluginconfigured = $plagiarismturnitin->is_plugin_configured();
+        $this->assertEquals(false, $ispluginconfigured);
+
+        // Check if plugin is configured with account id and apiurl set.
+        set_config('apiurl', 'http://www.test.com', 'turnitintooltwo');
+        $ispluginconfigured = $plagiarismturnitin->is_plugin_configured();
```
alex-bell: Seeing as you have to redeclare ispluginconfigured every time, would it be worth just whacking $plagiarismturnitin->is_plugin_configured() into the assert?

tests/lib_test.php
```diff
@@ -0,0 +1,61 @@
...
+        $this->assertEquals(false, $ispluginconfigured);
+
+        // Check if plugin is configured with account id, apiurl and secretkey set.
+        set_config('secretkey', 'ABCDEFGH', 'turnitintooltwo');
```
alex-bell: If you're testing the different elements of the if check in $plagiarismturnitin->is_plugin_configured(), don't you need to empty out the config so it doesn't just pass after the first one has been set, or a I misunderstanding what set_config does?

jmcgettrick: No, it's setting each individually and they carry over. So I'm testing when 1 is set, then when 2 are set then finally when they are all set.

tests/lib_test.php
jmcgettrick: I could, but I feel it reads better as it is.

PR #275, opened 2017-11-20
Fix for group submission grading

dwinn:
This pull request fixes the issue where grades for group submissions are not passed to the gradebook.


jmcgettrick:
Looks good to me

PR #277, opened 2017-11-23
Fix group submission grading when submitting on behalf of a student.

dwinn:
Fix group submission grading when submitting on behalf of a student.

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -700,7 +700,7 @@ class plagiarism_plugin_turnitin extends plagiarism_plugin {
                To get around this, we get the group ID, get the group members and set the author as the first student in the group.
             */
             $moodlesubmission = $DB->get_record('assign_submission', array('id' => $itemid), 'id, groupid');
-            if (!empty($moodlesubmission->groupid)) {
+            if ((!empty($moodlesubmission->groupid)) && ($cm->modname == "assign")) {
                 $author = $this->get_first_group_author($cm->course, $moodlesubmission->groupid);
                 $linkarray['userid'] = $author;
             } else {
@@ -1512,9 +1512,9 @@ class plagiarism_plugin_turnitin extends plagiarism_plugin {
         }
 
         $groupmembers = groups_get_members($groupid, "u.id");
-        foreach ($groupmembers as $author => $userobject) {
-            if (!has_capability('plagiarism/turnitin:grade', $context, $author)) {
-                return $author;
+        foreach ($groupmembers as $author) {
+            if (!has_capability('mod/assign:grade', $context, $author->id)) {
+                return $author->id;
             }
         }
     }
@@ -2308,7 +2308,7 @@ class plagiarism_plugin_turnitin extends plagiarism_plugin {
            Related user ID will be NULL if an instructor submits on behalf of a student who is in a group.
            To get around this, we get the group ID, get the group members and set the author as the first student in the group.
         */
-        if (empty($eventdata['relateduserid'])) {
+        if ((empty($eventdata['relateduserid'])) && ($eventdata['other']['modulename'] == 'assign')) {
             $moodlesubmission = $DB->get_record('assign_submission', array('id' => $eventdata['objectid']), 'id, groupid');
             if (!empty($moodlesubmission->groupid)) {
                 $author = $this->get_first_group_author($cm->course, $moodlesubmission->groupid);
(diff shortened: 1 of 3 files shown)


lib.php
```diff
@@ -695,12 +696,25 @@ public function get_links($linkarray) {
...
-                $author = $moduleobject->get_author($itemid);
-                $linkarray['userid'] = (!empty($author)) ? $author : $linkarray['userid'];
+            /*
+               The author will be incorrect if an instructor submits on behalf of a student who is in a group.
+               To get around this, we get the group ID, get the group members and set the author as the first student in the group.
+            */
+            $moodlesubmission = $DB->get_record('assign_submission', array('id' => $itemid), 'id, groupid');
+            if (!empty($moodlesubmission->groupid)) {
+                // Get the group members for the students in this group and set the author as the first member.
+                $groupmembers = current(groups_get_members_by_role($moodlesubmission->groupid, $cm->course, 'u.id', NULL, "r.id = 5"));
```
jmcgettrick: This should work most of the time but some institutions heavily customise the roles in Moodle so 5 might not necessarily be a student. It would probably be more reliable to loop thorugh the returned members from groups_get_members and use the first user who has the following permission (has_capability('mod/assign:grade', context_course::instance($cm->course))).

lib.php
```diff
@@ -2273,12 +2286,25 @@ public function event_handler($eventdata) {
...
 
+        /*
+           Related user ID will be NULL if an instructor submits on behalf of a student who is in a group.
+           To get around this, we get the group ID, get the group members and set the author as the first student in the group.
+        */
+        if (empty($eventdata['relateduserid'])) {
+            $moodlesubmission = $DB->get_record('assign_submission', array('id' => $eventdata['objectid']), 'id, groupid');
+            if ($moodlesubmission->groupid) {
+                // Get the group members for the students in this group and set the author as the first member.
+                $groupmembers = current(groups_get_members_by_role($moodlesubmission->groupid, $cm->course, 'u.id', null, "r.id = 5"));
```
jmcgettrick: See previous comment.

lib.php
```diff
@@ -695,12 +695,22 @@ public function get_links($linkarray) {
...
-            // post Moodle 2.7 which is problematic as teachers can submit on behalf of students.
-            $author = $linkarray['userid'];
-            if ($itemid != 0) {
-                $author = $moduleobject->get_author($itemid);
-                $linkarray['userid'] = (!empty($author)) ? $author : $linkarray['userid'];
+            /*
+               The author will be incorrect if an instructor submits on behalf of a student who is in a group.
+               To get around this, we get the group ID, get the group members and set the author as the first student in the group.
+            */
+            $moodlesubmission = $DB->get_record('assign_submission', array('id' => $itemid), 'id, groupid');
```
jmcgettrick: I think that needs to be within an if ($mod == assign) condition.

lib.php
```diff
@@ -1488,6 +1497,28 @@ public function check_group_submission($cm, $userid) {
...
+    */
+    private function get_first_group_author($cmid, $groupid) {
+        static $context;
+        if (empty($context)) {
+            $context = context_course::instance($cmid);
+        }
+
+        $groupmembers = groups_get_members($groupid, "u.id");
+        foreach ($groupmembers as $author => $userobject) {
+            if (!has_capability('plagiarism/turnitin:grade', $context, $author)) {
```
jmcgettrick: Personally, i would just use mod/assign:grade

lib.php
```diff
@@ -1488,6 +1497,28 @@ public function check_group_submission($cm, $userid) {
...
+     * @return int $author The Moodle user ID that we'll be using for the author.
+    */
+    private function get_first_group_author($cmid, $groupid) {
+        static $context;
+        if (empty($context)) {
+            $context = context_course::instance($cmid);
+        }
+
+        $groupmembers = groups_get_members($groupid, "u.id");
+        foreach ($groupmembers as $author => $userobject) {
```
jmcgettrick: Not a big deal but you don't need $userobject.

lib.php
```diff
@@ -2273,12 +2304,22 @@ public function event_handler($eventdata) {
         $submitter = $eventdata['userid'];
         $author = (!empty($eventdata['relateduserid'])) ? $eventdata['relateduserid'] : $eventdata['userid'];
 
+        /*
+           Related user ID will be NULL if an instructor submits on behalf of a student who is in a group.
+           To get around this, we get the group ID, get the group members and set the author as the first student in the group.
+        */
+        if (empty($eventdata['relateduserid'])) {
```
jmcgettrick: if && $mod == "assign"

Issue #284, opened 2017-12-15
Undefined variable notice


micaherne:
I'm seeing this message on the "view all submissions" page for an assignment:

> Notice: Undefined variable: notorlink in /var/www/html/plagiarism/turnitin/lib.php on line 906

dwinn:
Hi @micaherne,

This has been fixed in the latest release v2018011601. Thanks for the contribution.

PR #285, opened 2017-12-15
Fix undefined variable notice

micaherne:
This is a fix for the undefined variable notice in #284 

It doesn't change any functionality.


jmcgettrick:
Thanks @micaherne - I've added this in #286 so it goes to develop.

Issue #288, opened 2018-01-04
"Not authorised" error for group assignments


micaherne:
We are seeing the following error message for most group assignments after upgrading to 2017112302 (from 2017022201):

> Turnitin has returned an error with your submission: You are not authorized to access this resource.

This appears to be caused by the fix to #277, which is changing the author to be the first person in the group alphabetically. If this is not the person submitting the error occurs.

To reproduce:

1. Create an assignment. Enable group submission and Turnitin.
2. Create a group with 5 students.
3. Log in as the middle student (alphabetically by surname) and submit a document.
4. Run the send_submissions task
5. Go to the "view all submissions" page where you should see the error.

micaherne:
Just to note, we've checked that the fix for #277 is working fine - this is only happening for student group submissions.

micaherne:
Also, this is on Moodle 3.2.6 (Build: 20171113).

dwinn:
Hi @micaherne,

This has been fixed in the latest release v2018011601. Thanks for the contribution.

PR #289, opened 2018-01-05
Fix group submissions

micaherne:
Hi, here is a patch for https://github.com/turnitin/moodle-plagiarism_turnitin/issues/288

All it does is:

1. Only saves a different submitter from the author when the submission is probably being edited by a tutor, rather than submitted by a student.
2. Looks up the user who made the latest group submission explicitly in get_links(), rather than inferring it from the group membership.

I've tested it pretty thoroughly and as far as I can see it's sound, but it would be good if someone could confirm that there are no scenarios where the queries on L705 and L811 could return a different record.

There's one outstanding edge case that it doesn't cover, which is where someone has the 'mod/assign:editothersubmission' capability on an assignment in Moodle, but Turnitin thinks they're a student. I think this scenario is probably technically possible, but I can't see why anyone would set their permissions up this way. In any case I don't think there's enough information being passed into the event handler to deal with this.

Hope it's useful.

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -699,9 +699,13 @@ class plagiarism_plugin_turnitin extends plagiarism_plugin {
                The author will be incorrect if an instructor submits on behalf of a student who is in a group.
                To get around this, we get the group ID, get the group members and set the author as the first student in the group.
             */
+            $plagiarismfile = null;
             $moodlesubmission = $DB->get_record('assign_submission', array('id' => $itemid), 'id, groupid');
             if ((!empty($moodlesubmission->groupid)) && ($cm->modname == "assign")) {
-                $author = $this->get_first_group_author($cm->course, $moodlesubmission->groupid);
+                $plagiarismfiles = $DB->get_records('plagiarism_turnitin_files', ['itemid' => $itemid, 'cm' => $cm->id, 'identifier' => $identifier],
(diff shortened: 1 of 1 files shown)


dwinn:
Hi @micaherne,

Thank you for the contribution. I have tested your branch and it looks good. It fixes the issue you reported. I agree the edge case you mention would be difficult to fix, I think we will leave that for now and look into it later if need be rather than hold this back until we can find a way to do that.

One thing though. Your pull request is to master branch but our release process requires new code to be merged into develop first. Would you be able to make a new pull request to develop?

Thanks.

micaherne:
Hi @dwinn, thanks for reviewing the patch. I've redone it based on the develop branche, so hopefully that'll work. Cheers, Michael.

dwinn:
Hi @micaherne,

Thanks. We'll get this out in the upcoming release. We can't give a date as we don't want to rush the testing but we aim to release around the middle of January.

micaherne:
Thanks @dwinn. Are you planning to do anything in upgrade.php to fix the broken data?

dwinn:
Hi @micaherne,

We will need to do some investigation work to see whether that is possible before we can do that and ensure it can't alter data that shouldn't be altered. For now the fix will go out without this.

micaherne:
Thanks @dwinn, that's good to know.

PR #290, opened 2018-01-08
Support report gen speed changes in Moodle PP

dwinn:
Support report gen speed changes in Moodle PP


turnitinplugin_view.class.php
```diff
@@ -134,8 +134,9 @@ public function add_elements_to_settings_form($mform, $course, $location = "acti
         }
 
         $options = array(0 => get_string('no'), 1 => get_string('yes'));
+        $genparams = turnitintooltwo_get_report_gen_speed_params();
```
jmcgettrick: Should that not be test_plagiarism_get_report_gen_speed_params(). I know it'll still work as they're from v2 and the values are the same.

jmcgettrick:
Looks fine apart from the one method call. 👍

turnitinplugin_view.class.php
dwinn: Copy and paste fail and I missed it since it's linked to V2 and worked.

PR #303, opened 2018-04-19
Remove setting dependencies

dwinn:
Remove setting dependencies.

Created a Moodle form object to replicate V2 settings in PP.
Copied the connection test functionality across.
Copied all language strings that are required for the settings pages.

As of now, there are still dependencies on V2 being setup for class creation and user creation, so this needs to be kept in mind for QA.


styles.css
```diff
@@ -732,4 +732,21 @@ html .tooltipster-base {
 
 .selectlong {
     width: 530px;
+}
+
+.connection-test-success,
+.connection-test-failed {
+    background-image: none !important;
```
alex-bell: are the importants necessary?

ip-pauldawson:
This is the sort of refactor where you'd be really glad you have good tests for 😆 

(This needs to be tested a lot)

classes/forms/tiisetupform.class.php
```diff
@@ -0,0 +1,227 @@
...
+
+class tiisetupform extends moodleform {
+
+    // Define the form.
+    public function definition() {
+        global $DB;
+
+        $config = plagiarism_plugin_turnitin::plagiarism_turnitin_admin_config();
+
+        $mform =& $this->_form;
```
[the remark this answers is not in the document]
dwinn: Looking at it again I don't think we need the reference at all. The Moodle form API example is to just use:

```$mform = $this->_form;```

I had the `&` as I used another form in our plugin as the class template for starting.

File: styles.css (hunk 1 above)
Comment (dwinn) [reply styles.css]: Turns out no.

File: classes/forms/tiisetupform.class.php [hunk 3]
```diff
@@ -0,0 +1,227 @@
...
+
+        $mform->addElement('select', 'plagiarism_turnitin_useerater', get_string('turnitinuseerater', 'plagiarism_turnitin'), $ynoptions);
+        $mform->addElement('static', 'plagiarism_turnitin_useerater_desc', null, get_string('turnitinuseerater_desc', 'plagiarism_turnitin'));
+        $mform->setDefault('plagiarism_turnitin_useerater', 0);
+
+        $mform->addElement('select', 'plagiarism_turnitin_transmatch', get_string('transmatch', 'plagiarism_turnitin'), $ynoptions);
+        $mform->addElement('static', 'plagiarism_turnitin_transmatch_desc', null, get_string('transmatch_desc', 'plagiarism_turnitin'));
+        $mform->setDefault('plagiarism_turnitin_transmatch', 0);
+
+        $repositoryoptions = array(
```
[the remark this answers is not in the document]
dwinn: I've added it.

lang/cs/plagiarism_turnitin.php
```diff
@@ -224,9 +224,9 @@
 $string['turnitinrepositoryoptions'] = 'Úkoly archivu prací';
 $string['turnitinrepositoryoptions_desc'] = 'Zvolte možnosti archivu pro úkoly Turnitin. <br /><i>(Archiv instituce mají k dispozici pouze ti, kdo si ho aktivovali u svého účtu)</i>';
 $string['tiimiscsettings'] = 'Různé nastavení pluginu';
-$string['turnitintooltwoagreement_default'] = 'Potvrzuji, že tato odevzdaná práce je moje vlastní práce, a přijímám veškerou zodpovědnost za případná porušení autorských práv, která mohou nastat v důsledku tohoto odevzdání.';
-$string['turnitintooltwoagreement_desc'] = '<b>[Optional]</b><br />Zadejte prohlášení o potvrzení dohody pro odevzdání prací.<br />(<b>Poznámka:</b> Pokud je dohoda ponechána zcela prázdná, nebude od studentů během odevzdání požadováno její potvrzení)';
-$string['turnitintooltwoagreement'] = 'Prohlášení / Dohoda';
+$string['ppagreement_default'] = 'Potvrzuji, že tato odevzdaná práce je moje vlastní práce, a přijímám veškerou zodpovědnost za případná porušení autorských práv, která mohou nastat v důsledku tohoto odevzdání.';
```
alex-bell: Would it be clearer for these prefixes to be `pp_` ?

PR #305, opened 2018-04-25
Moodle 35 fixes

dwinn:
Fixed an issue where send_queued_files task would fail.
Changed depecated stdClass() to new \core\message\message() for event data.
Fixed an issue where the $cm parameter was not being passed in during a text submission.

### DIFF
--- a/classes/digitalreceipt/pp_receipt_message.php
+++ b/classes/digitalreceipt/pp_receipt_message.php
@@ -31,11 +31,17 @@ class pp_receipt_message {
      * @param string $message
      * @return void
      */
-    public function send_message($userid, $message) {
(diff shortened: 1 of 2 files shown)


classes/digitalreceipt/pp_receipt_message.php
```diff
@@ -47,6 +53,10 @@ public function send_message($userid, $message) {
         $eventdata->smallmessage      = '';
         $eventdata->notification      = 1; // This is only set to 0 for personal messages between users.
 
+        if ($CFG->branch >= 32) {
+            $eventdata->courseid = $courseid;
+        }
```
jmcgettrick: Does adding the courseid break pre 3.2?

PR #309, opened 2018-05-21
GDPR implementation

dwinn:
GDPR implementation


danmarsden:
nice - I might borrow some of this :-)

should "link_external_location" include information to say that the assignment file is also sent externally? - (I haven't tested this to see how it displays)

PR #316, opened 2018-05-25
Remove class dependency from PP

dwinn:
Note: Until the assignment logic is moved across then this will not remove all dependencies.

This is because when a new course is created in Moodle, it will only be created in the database tables when the first assignment/submission is made and the cron runs.

What this means is that the cron will call `get_course_data`, which in turn calls `create_tii_course` if a course doesn't exist. This then calls a function of the same name in `turnitintooltwo_assignment class` which inserts into `turnitintooltwo_courses`.

What does change in this PR is we now have a courses table with the data migrated across during a plugin upgrade, and all references to `turnitintooltwo_courses` have been removed with the new `plagiarism_turnitin_courses` table.

### DIFF
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -331,7 +331,6 @@ function xmldb_plagiarism_turnitin_upgrade($oldversion) {
                     $course->turnitin_ctl = $ppcourse->turnitin_ctl;
                     $DB->insert_record('plagiarism_turnitin_courses', $course);
 
-
                     // Clean up the record from the V2 plugin.
                     $coursev2 = array('courseid' => $ppcourse->courseid,
                         'ownerid' => $ppcourse->ownerid,
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -298,6 +298,27 @@ function xmldb_plagiarism_turnitin_upgrade($oldversion) {
         }
     }
 
+    if ($oldversion < 2018052402) {
+        // Define table file_conversion to be created.
+        $table = new xmldb_table('plagiarism_turnitin_courses');
+
+        // Adding fields to table plagiarism_turnitin_courses.
+        $table->add_field('id', XMLDB_TYPE_INTEGER, '10', null, XMLDB_NOTNULL, XMLDB_SEQUENCE, null);
+        $table->add_field('courseid', XMLDB_TYPE_INTEGER, '10', null, false, null, null, 'id');
+        $table->add_field('ownerid', XMLDB_TYPE_INTEGER, '10', null, false, null, null, 'courseid');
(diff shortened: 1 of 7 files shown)


db/upgrade.php
```diff
@@ -298,6 +298,50 @@ function xmldb_plagiarism_turnitin_upgrade($oldversion) {
...
+            $ppcourses = $DB->get_records('turnitintooltwo_courses', array('course_type' => 'PP'), 'id ASC', 'courseid, ownerid, turnitin_ctl, turnitin_cid');
+
+            foreach ($ppcourses as $ppcourse) {
+                $course = new stdClass();
+                $course->courseid = $ppcourse->courseid;
+                $course->ownerid = $ppcourse->ownerid;
+                $course->turnitin_cid = $ppcourse->turnitin_cid;
+                if (!$DB->get_record('plagiarism_turnitin_courses', array('courseid' => $ppcourse->courseid))) {
+                    $course->turnitin_ctl = $ppcourse->turnitin_ctl;
+                    $DB->insert_record('plagiarism_turnitin_courses', $course);
```
jmcgettrick: IIRC, this can't be done in bulk can it?

dwinn: There is an `insert_records()` method in Moodle but this essentially does the same thing - a loop calling `insert_record()`. I know that in SQL it is possible to perform a multi-insert in one query but I can't see any method to do this via Moodle DB API and I'd rather we stuck with that otherwise we'll have to scale test it across all DBs.

We'd also lose the check for whether the record already exists if we do a multi-insert - though I see no reason why a record should exist at this point. Perhaps only if a version upgrade were to fail and had to be re-ran. Duplicates can be prevented via an index too I suppose, if we handle the error returned.

Happy to rewrite it though if you can think of a better way.

danmarsden: FYI - insert_records does do stuff in bulk rather than iterating over each and doing a single insert_record (I use insert_records a lot and find it really efficient.)

it does depend on which db type and O/S is in use on how efficient it really is. Postgres db does 300 records at a time on a single insert.

mysql on windows does 5 at a time, mysql on other o/s does a check on max_allowed_packet and then sets an appropriate number of records to do at a time up to a max hard-coded limit of 50 records in one go.

danmarsden: - MS SQL/Oracle use the default insert_records() function which does just iterate over insert_record()

dwinn: Good info @danmarsden. I think I may see if I can rework it to use the `inserts_records` function, it might make it smoother for larger institutions.

dwinn: I've made some changes to use `insert_records`

jmcgettrick:
Looks good, just need to move the code out of the view class.

dwinn:
New pull request is #323.

PR #319, opened 2018-05-30
Add peermark launch to turnitin_view

ip-pauldawson:
- Add PeerMark and QM launches to turnitin_view class

### DIFF
--- a/classes/turnitin_view.class.php
+++ b/classes/turnitin_view.class.php
@@ -551,4 +551,65 @@ class turnitin_view {
             return $turnitincall->$ltifunction($lti, $ltireturn);
         }
     }
+
+    /**
+     * Return the output for a form to launch the relevant LTi function
+     * It is then submitted on load via Javascript
+     *
+     * @param string $userrole either Instructor or Learner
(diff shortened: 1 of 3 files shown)


classes/turnitin_view.class.php
```diff
@@ -551,4 +551,65 @@ public static function output_launch_form($type, $submissionid, $userid, $userro
...
+     * It is then submitted on load via Javascript
+     *
+     * @param string $userrole either Instructor or Learner
+     * @param int $userid
+     * @return output form
+     */
+    public static function output_lti_form_launch($type, $userrole, $partid = 0, $classid = 0) {
+        global $USER;
+        // Initialise Comms Object.
+        $turnitincomms = new turnitintooltwo_comms();
```
dwinn: I think we should change this to use `turnitin_comms`. The comms stuff has been moved over, although I see a second (unrelated) reference to `turnitintooltwo_comms` in ajax.php as well. Up to you if you wish to change the second one but might explain why the test connection isn't working for me. We can make a separate ticket for the latter if you prefer.

PR #323, opened 2018-05-31
Feature remove class dependency

dwinn:
This pull request replaces the broken #316.

Note: Until the assignment logic is moved across then this will not remove all dependencies.

This is because when a new course is created in Moodle, it will only be created in the database tables when the first assignment/submission is made and the cron runs.

What this means is that the cron will call get_course_data, which in turn calls create_tii_course if a course doesn't exist. This then calls a function of the same name in turnitintooltwo_assignment class which inserts into turnitintooltwo_courses.

What does change in this PR is we now have a courses table with the data migrated across during a plugin upgrade, and all references to turnitintooltwo_courses have been removed with the new plagiarism_turnitin_courses table.

course_type field is removed in the new table as it is no longer required, however it remains in the V2 table as I believe the migration tool uses it.

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -409,7 +409,7 @@ class plagiarism_plugin_turnitin extends plagiarism_plugin {
         // If all turnitin enabled modules for this course have been reset.
         // then remove the Turnitin course id from the database.
         if ($resetcourse) {
-            $DB->delete_records('plagiarism_turnitin_courses', array('courseid' => $courseid, 'course_type' => 'PP'));
+            $DB->delete_records('plagiarism_turnitin_courses', array('courseid' => $courseid));
         }
 
         return true;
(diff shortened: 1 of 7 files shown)


lib.php
```diff
@@ -409,7 +409,7 @@ public static function course_reset($eventdata) {
         // If all turnitin enabled modules for this course have been reset.
         // then remove the Turnitin course id from the database.
         if ($resetcourse) {
-            $DB->delete_records('turnitintooltwo_courses', array('courseid' => $courseid, 'course_type' => 'PP'));
+            $DB->delete_records('plagiarism_turnitin_courses', array('courseid' => $courseid, 'course_type' => 'PP'));
```
ip-pauldawson: remove `course_type`

ip-pauldawson:
Looks good

PR #324, opened 2018-05-31
Fix default settings errors and rename the plugin settings form.

dwinn:
Fix errors when accessing default settings due to the relocation of the view class. It was being called by a class in Moodle Direct v2 - This dependency has been removed but the redundant code can't be removed from V2 until we release this.

Also renamed the setup form to be consistent with other classes.


ip-pauldawson:
Looks good now

PR #332, opened 2018-06-11
Copy assignment class to PP

dwinn:
This pull request copies the assignment.class methods from Moodle Direct V2 that will be required for the Plagiarism Plugin to work independently.

- All calls to turnitintooltwo_assignment have been replaced with the new PP class.
- Methods have been updated for PP specific functionality.
- Redundant code has been removed (eg checks for course type)
- Language strings and translations have been moved.
- Unit tests have been added for all methods - these are new and were not currently in MD V2.
- Travis file has been updated to check more versions - this will require further refining once the split has been completed.


tests/classes/turnitin_assignment_class_test.php
```diff
@@ -0,0 +1,317 @@
...
+        $course->turnitin_cid = 10;
+        $course->fullname = "This is a test course";
+        $course->tii_rel_id = 1234;
+
+        // Insert the course to the plagiarism turnitin courses table.
+        $course->id = $DB->insert_record('plagiarism_turnitin_courses', $course);
+
+        // Stub a fake tii comms.
+        $faketiicomms = $this->getMockBuilder(turnitin_comms::class)
+            ->disableOriginalConstructor()
```
jmcgettrick: 👌

.travis.yml
```diff
@@ -16,12 +16,19 @@ env:
   - IGNORE_PATHS=sdk/,pix/
   - IGNORE_NAMES=jquery*,styles.css
  matrix:
-  # - DB=pgsql MOODLE_BRANCH=MOODLE_31_STABLE
-  # - DB=mysqli MOODLE_BRANCH=MOODLE_31_STABLE
+#  - DB=pgsql MOODLE_BRANCH=MOODLE_32_STABLE
```
jmcgettrick: Shall we update this? Put 3.5 in?

jmcgettrick: Just a general observation really

dwinn: Yeah we should. I'll stick it in the PR I'll be making for PP develop for my other ticket and then it will make its way into the split branch.

PR #333, opened 2018-06-14
Feature - Add behat tests to Turnitin Plagiarism Plugin

ip-pauldawson:
- Add behat smoke and sanity tests
- Add custom step definitions


tests/behat/installed.feature
```diff
@@ -0,0 +1,33 @@
...
+  I need the installation to work and plagiarism plugins to be enabled
+
+  Background: Set up the plugin
+    When I log in as "admin"
+    And I navigate to "Advanced features" node in "Site administration"
+    And I set the field "Enable plagiarism plugins" to "1"
+    And I press "Save changes"
+    And I navigate to "Turnitin" node in "Site administration > Plugins > Plagiarism"
+    And I set the following fields to these values:
+      | Enable Turnitin            | 1 |
```
alex-bell: Can something reusable like this be part of a setup stage?

jmcgettrick: Afraid not, the background step it's part of is the setup

ip-pauldawson: I'll be creating some additional tickets to expand coverage, we could possibly create a custom definition to reduce duplication maybe, combining some of these repeated steps.

alex-bell: 👍

PR #334, opened 2018-06-15
Fix duplicate submissions in PP Turnitin

dwinn:
Fix duplicate submissions in PP Turnitin

### DIFF
--- a/classes/modules/turnitin_assign.class.php
+++ b/classes/modules/turnitin_assign.class.php
@@ -93,12 +93,13 @@ class turnitin_assign {
             if (($attemptreopenmethod == ASSIGN_ATTEMPT_REOPEN_METHOD_NONE) ||
                 ($attemptreopened == "submitted")) {
                 // If this is a text or file submission, or we can only submit one file.
-                if ($submissiontype == 'text_content' || $submissiontype == 'file' || $maxfilesubmissions == 1) {
+                if ($submissiontype == 'text_content' || ($submissiontype == 'file' && $maxfilesubmissions == 1)) {
                     // Treat this as a resubmission.
                     return true;
                 }
             }
         }
+        return false;
     }
 
     public function get_onlinetext($userid, $cm) {
--- a/classes/modules/turnitin_assign.class.php
+++ b/classes/modules/turnitin_assign.class.php
@@ -68,7 +68,7 @@ class turnitin_assign {
      *
      * @param $assignid
      */
-    public function is_resubmission_allowed($assignid, $reportgenspeed, $submissiontype, $attemptreopenmethod) {
+    public function is_resubmission_allowed($assignid, $reportgenspeed, $submissiontype, $attemptreopenmethod, $attemptreopened = null) {
         global $DB, $CFG;
 
         // Get the maximum number of file submissions allowed.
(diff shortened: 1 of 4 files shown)


classes/modules/turnitin_assign.class.php
```diff
@@ -85,9 +85,21 @@ public function is_resubmission_allowed($assignid, $reportgenspeed, $submissiont
         if ($CFG->branch <= 32) {
             require_once($CFG->dirroot.'/mod/assign/lib.php');
         }
-
-        return ($reportgenspeed > 0 && $attemptreopenmethod == ASSIGN_ATTEMPT_REOPEN_METHOD_NONE
-            && ($submissiontype == 'text_content' || $maxfilesubmissions == 1));
+        require_once($CFG->dirroot . '/mod/assign/locallib.php');
```
jmcgettrick: Do we need lib.php above as well?

lib.php
```diff
@@ -2141,13 +2141,31 @@ public function queue_submission_to_turnitin($cm, $author, $submitter, $identifi
         // Get module data.
         $moduledata = $DB->get_record($cm->modname, array('id' => $cm->instance));
         $moduledata->resubmission_allowed = false;
+
         if ($cm->modname == 'assign') {
+            if (!isset($_SESSION["moodlesubmissionstatus"])) {
```
jmcgettrick: Why is that in the `$_SESSION`?

dwinn: It's to deal with attempts. During a text submission `queue_submission_to_turnitin` is called twice. The first time this is ran the submission status is `reopened`, which is the only way I could determine that this is the first submission of a new attempt. The second time it is ran this has already been changed to `submitted`. It can't be stored as a variable because it wouldn't be there for the next call to `queue_submission_to_turnitin()`. I'm not sure why we call this twice.

classes/modules/turnitin_assign.class.php
```diff
@@ -82,12 +82,21 @@ public function is_resubmission_allowed($assignid, $reportgenspeed, $submissiont
             $maxfilesubmissions = $result->value;
         }
 
-        if ($CFG->branch <= 32) {
-            require_once($CFG->dirroot.'/mod/assign/lib.php');
+        require_once($CFG->dirroot . '/mod/assign/locallib.php');
+
+        // If resubmissions are enabled in a Turnitin sense.
+        if ($reportgenspeed > 0) {
+            // If the attempt reopened methos is none, or an attempt has not been reopened/has previous submission.
```
alex-bell: method*

classes/modules/turnitin_assign.class.php
```diff
@@ -82,12 +82,21 @@ public function is_resubmission_allowed($assignid, $reportgenspeed, $submissiont
...
+            if (($attemptreopenmethod == ASSIGN_ATTEMPT_REOPEN_METHOD_NONE) ||
```
alex-bell: Does this need those inner parentheses?

classes/modules/turnitin_assign.class.php
```diff
@@ -82,12 +82,21 @@ public function is_resubmission_allowed($assignid, $reportgenspeed, $submissiont
...
+                ($attemptreopened == "submitted")) {
```
alex-bell: Does this need double quotes rather than single?

lib.php
```diff
@@ -2141,13 +2141,31 @@ public function queue_submission_to_turnitin($cm, $author, $submitter, $identifi
...
+                $_SESSION["moodlesubmissionstatus"] = null;
```
alex-bell: If it's not set then isn't it already null?

alex-bell: If it's not set, then isn't already null?

classes/modules/turnitin_assign.class.php
alex-bell: Does this need to be double quotes rather than single

classes/modules/turnitin_assign.class.php
alex-bell: Does this need the inner parentheses?

classes/modules/turnitin_assign.class.php
alex-bell: Method*

lib.php
```diff
@@ -2141,13 +2141,31 @@ public function queue_submission_to_turnitin($cm, $author, $submitter, $identifi
...
+            }
+
+            if (($eventtype == "content_uploaded") || ($eventtype == "file_uploaded")) {
```
alex-bell: Does this need the inner parentheses?

lib.php
```diff
@@ -2141,13 +2141,31 @@ public function queue_submission_to_turnitin($cm, $author, $submitter, $identifi
...
             $moduledata->resubmission_allowed = $moduleobject->is_resubmission_allowed(
                 $cm->instance,
                 $settings["plagiarism_report_gen"],
                 $submissiontype,
-                $moduledata->attemptreopenmethod
+                $moduledata->attemptreopenmethod,
+                $_SESSION["moodlesubmissionstatus"]
             );
+
+            if (($eventtype != "content_uploaded") && ($eventtype != "file_uploaded")) {
```
alex-bell: Does this need the inner parentheses?

classes/modules/turnitin_assign.class.php
dwinn: You can use either in PHP but I'll change for consistency.

lib.php
dwinn: This is here so we can set the variable and give it a default value if it hasn't been set already. Otherwise the code might give an undefined index warning further down when passed to `is_resubmission_allowed()`

PR #338, opened 2018-06-29
Remove user dependency

dwinn:
This pull request removes the PP dependency on the Moodle Direct V2 user class and its associated database table. The unlink/relink tool has also been imported into the plugin.


classes/turnitin_user.class.php
```diff
@@ -0,0 +1,545 @@
...
+        $this->firstname = "";
+        $this->lastname = "";
+        $this->fullname = "";
+        $this->email = "";
+        $this->username = "";
+
+        if ($id != 0) {
+            $this->get_moodle_user($this->id);
+            if ($finduser === true) {
+                echo 'aaa';
```
jmcgettrick: aaa

classes/turnitin_user.class.php
```diff
@@ -0,0 +1,545 @@
...
+     * @return string A pseudo lastname address
+     */
+    public function get_pseudo_lastname() {
+        global $DB;
+        $config = plagiarism_plugin_turnitin::plagiarism_turnitin_admin_config();
+        $userinfo = $DB->get_record('user_info_data', array('userid' => $this->id, 'fieldid' => $config->pseudolastname));
+
+        if ((!isset($userinfo->data) || empty($userinfo->data)) && $config->pseudolastname != 0 && $config->lastnamegen == 1) {
+            $uniqueid = strtoupper(strrev(uniqid()));
+            $userinfoob = new stdClass();
```
jmcgettrick: `$userinfoob` - made me 😂 for some reason

classes/turnitin_user.class.php
```diff
@@ -0,0 +1,545 @@
...
+
+
+    /**
+     * Check Turnitin to see if this user has previously registered
+     *
+     * @param object $user_details A data object for the user
+     * @return var Turnitin user id if found otherwise null
+     */
+    private function find_tii_user_id() {
+        echo 'aaa';
```
jmcgettrick: aaa

classes/turnitin_user.class.php
```diff
@@ -0,0 +1,543 @@
...
+            $finduser = $response->getUser();
+            $tiiuserid = $finduser->getUserId();
+            if (empty($tiiuserid)) {
+                $tiiuserid = null;
+            }
+            return $tiiuserid;
+
+        } catch (Exception $e) {
+            $toscreen = ($this->workflowcontext == "cron") ? false : true;
+            $this->turnitincomms->handle_exceptions($e, 'userfinderror', $toscreen);
```
jmcgettrick: I think that `$this->workflowcontext` won't be needed for PP but I'm not certain. I think it was needed when this class was in v2. Leave as is but it's something to bear in mind for the future that we might be able to remove it.

classes/forms/turnitin_unlinkform.class.php
```diff
@@ -0,0 +1,144 @@
...
+require_once($CFG->libdir."/formslib.php");
+
+class turnitin_unlinkform extends moodleform {
+
+    // Define the form.
+    public function definition()
+    {
+        $mform =& $this->_form;
+
+        foreach ($this->_customdata["elements"] as $element) {
```
hperperidis: This loop could probably be refactored to something more readable. E.g. call different methods for the type of element you want to add on the form.

But if out of the scope of this ticket, just ignore this comment.

classes/forms/turnitin_unlinkform.class.php
```diff
@@ -0,0 +1,144 @@
...
+                case "text":
+                    $mform->addElement($element[0], $element[1], $element[2]);
+                    $mform->setType($element[1], PARAM_RAW);
+                    break;
+                default:
+                    $mform->addElement($element[0], $element[1], $element[2]);
+                    break;
+            }
+
+            // Set form data. Only used for submission form.
```
hperperidis: if this is only used for the submission form, is this needed here?

dwinn: From what I can see, this is only used in V2 so removing.

classes/forms/turnitin_unlinkform.class.php
dwinn: I think this is outside the scope of the ticket but definitely needs doing. We'll have a chat about it once you're back in.

PR #339, opened 2018-07-02
Ensure the PP user class is called and not V2's

dwinn:
Ensure the PP user class is called and not V2's.

Also fixes the call to the comms class. (as a side note it is correct that this calls V2 as moving that across is a separate ticket and needs to be done last)


classes/turnitin_view.class.php
```diff
@@ -20,6 +20,7 @@
 
 require_once($CFG->dirroot.'/mod/turnitintooltwo/turnitintooltwo_form.class.php');
```
hperperidis: is this still used? does it need to be removed?

dwinn: This is no longer used, removing.

dwinn:
Merging as Harry has given approval and doesn't have merge rights.

PR #344, opened 2018-07-25
Remove V2 SDK dependency and add logging

dwinn:
This pull request implements changes necessary to use the SDK from a vendor directory/composer as opposed to using V2's SDK. There will exist a separate pull request for removing the unused SDK and adding the new vendor directory as the number of lines changed would make this difficult to review.

This pull request also implements activity logs and API logs.
Some of the setting parameter names were also updated with the correct names such as the pseudo values.

This pull request has a dependency on a pull request to the SDK, and as such this should not be merged before the other as I will need to update the composer file with the correct version of the SDK. It will also have a dependency on the vendor directory that will exist in a new pull request.


lib.php
```diff
@@ -75,9 +79,6 @@
 // Get helper methods.
 require_once(__DIR__.'/locallib.php');
 
-// This can't be removed until comms and activity logs are copied across.
-require_once($CFG->dirroot.'/mod/turnitintooltwo/lib.php');
-
 // Include plugin classes.
 require_once(__DIR__.'/classes/turnitin_assignment.class.php');
```
ip-pauldawson: Maybe change this to be consistent with the bulk of the plugin and use `$CFG->dirroot` going forward.

PR #350, opened 2018-07-30
Fix behat tests and add missing string

dwinn:
This pull request fixes the behat tests for this branch. The tests were built as part of develop and were calling V2 settings however this plugin has its own settings now that needed to be called.

Also fixed a findUser issue that was created as a result of the new SDK implementation, and added a language string which was missing for an error message.

### DIFF
--- a/classes/turnitin_user.class.php
+++ b/classes/turnitin_user.class.php
@@ -17,6 +17,7 @@
 use Integrations\PhpSdk\TiiUser;
 use Integrations\PhpSdk\TiiPseudoUser;
 use Integrations\PhpSdk\TiiMembership;
+use Integrations\PhpSdk\TurnitinApiException;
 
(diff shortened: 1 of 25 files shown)


classes/turnitin_user.class.php
```diff
@@ -173,10 +173,12 @@ private function get_tii_user_id() {
         }
 
         if (empty($this->tiiuserid)) {
-            $this->tiiuserid = $this->find_tii_user_id();
-            if (empty($this->tiiuserid) && $this->enrol) {
+            try {
+                $this->tiiuserid = $this->find_tii_user_id();
+            } catch (Exception $e) {
```
ip-pauldawson: Should catch `TurnitinApiException` and let Moodle handle any other exception here.

PR #354, opened 2018-08-03
Misc tasks for finalising split

dwinn:
Moved EULA form across.
Moved any remaining JS across.
Moved the rubric handling across.
Moved the Quickmark handling across.
Renamed pp_turnitin_ula class.
Removed V2 version dependency. This branch now works without V2 installed.


jquery/turnitin_module.js
```diff
@@ -157,16 +158,16 @@ jQuery(document).ready(function($) {
     }
 
     // Hide the submission form if the user has never accepted or declined the Turnitin EULA.
-    if ($(".pp_turnitin_ula_ignored").length > 0) {
```
ip-pauldawson: ew, nice catch

PR #356, opened 2018-08-17
Int 8576 cron gets stuck if report scores cannot be updated

hperperidis:


lang/en/plagiarism_turnitin.php
```diff
@@ -224,3 +224,4 @@
 $string['privacy:metadata:plagiarism_turnitin_client:submission_filename'] = 'The name of the submitted file is sent to Turntin so that it is identifiable.';
 
 $string['privacy:metadata:core_files'] = 'Turnitin Assignment stores files that have been uploaded to Moodle to form a Turnitin submission.';
+$string['errorcode13'] = 'This submissionid - {$a->externalid} was not found in TII. Unable to retrieve similarity score and other submission data.';
```
dwinn: I'd change TII to Turnitin.

lib.php
```diff
@@ -2032,6 +2060,55 @@ public function cron_update_scores($submissiontype = 'file', $submissions) {
         return true;
     }
 
+    private function checklocalsubmissionstate($assignmentIds, $submissionids) {
```
dwinn: Methods can have underscores in Moodle so I'd change this to `change_local_submission_state()` for readability.

lib.php
```diff
@@ -2032,6 +2060,55 @@ public function cron_update_scores($submissiontype = 'file', $submissions) {
...
+        // Initialise Comms Object.
+        $turnitincomms = new turnitin_comms();
+        $turnitincall = $turnitincomms->initialise_api();
+        $tiisubmissionids = array();
+
+        foreach ($assignmentIds as $assignmentId) {
+            $submission = new TiiSubmission();
+            $submission->setAssignmentId($assignmentId);
+            $submission->setDateFrom('2018-09-12T09:00:00Z'); // TODO(hperperidis): see how we can get a valid date here;
```
dwinn: This looks like a work in progress. I think hardcoding the date like this could cause issues.

lib.php
```diff
@@ -2032,6 +2060,55 @@ public function cron_update_scores($submissiontype = 'file', $submissions) {
...
+            }
+        }
+
+        return array(
+            'trimmedSubmissions' => array_intersect($submissionids, $tiisubmissionids),
+            'missingTiiSubmissions' => array_diff($submissionids, $tiisubmissionids)
+        );
+    }
+
+    private function invalidatemissingsubmission($missingsubmission) {
```
dwinn: Similarly a better name would be `invalidate_missing_submission`.

lib.php
```diff
@@ -2032,6 +2060,55 @@ public function cron_update_scores($submissiontype = 'file', $submissions) {
...
+            'id, 
+            cm, 
+            externalid, 
+            userid'
+        );
+        $plagiarismfile = new stdClass();
+        $plagiarismfile->id = $currentsubmission->id;
+        $plagiarismfile->externalid = $currentsubmission->externalid;
+        $plagiarismfile->userid = $currentsubmission->userid;
+        $plagiarismfile->statuscode='error';
```
dwinn: Some spacing needed around the `=`

lib.php
```diff
@@ -1952,14 +1954,41 @@ public function cron_update_scores($submissiontype = 'file', $submissions) {
...
+                    if (!isset(array_flip($assignmentids)[$moduleturnitinconfig->value])) {
+                        $assignmentids[] = $moduleturnitinconfig->value;
+                    }
                 }
             }
         }
 
-        if (count($submissionids) > 0) {
+        //Seperate out $submissionids that do not exist in TII.
+        $validatedSubmissions = $this->checklocalsubmissionstate($assignmentids, $submissionids);
```
dwinn: Agree with Clegg, we can't use variables with camel case as they fail Travis check. Variables should be all lowercase and no underscores. Not something I personally agree with but we have to follow the Moodle standards.

lib.php
```diff
@@ -2032,6 +2060,55 @@ public function cron_update_scores($submissiontype = 'file', $submissions) {
         return true;
     }
 
+    private function checklocalsubmissionstate($assignmentIds, $submissionids) {
+        // Initialise Comms Object.
+        $turnitincomms = new turnitin_comms();
+        $turnitincall = $turnitincomms->initialise_api();
+        $tiisubmissionids = array();
+
+        foreach ($assignmentIds as $assignmentId) {
```
dwinn: Camels

lang/en/plagiarism_turnitin.php
hperperidis: 👍

lib.php
hperperidis: 👍

lib.php
```diff
@@ -1970,10 +1970,10 @@ public function cron_update_scores($submissiontype = 'file', $submissions) {
             }
         }
 
-        //Seperate out $submissionids that do not exist in TII.
+        // Seperate out $submissionids that do not exist in TII.
```
dwinn: `Separate`

lib.php
```diff
@@ -2060,21 +2060,21 @@ public function cron_update_scores($submissiontype = 'file', $submissions) {
         return true;
     }
 
-    private function check_local_submission_state($assignmentIds, $submissionids) {
+    private function check_local_submission_state($assignmentids, $submissionids) {
```
dwinn: Need to rename this function to match the call to it without underscores.

PR #357, opened 2018-08-20
Fix rubric option so that rubrics are saved - split branch.

dwinn:
Fix rubric option so that rubrics are saved - split branch.

### DIFF
--- a/classes/turnitin_view.class.php
+++ b/classes/turnitin_view.class.php
@@ -122,7 +122,7 @@ class turnitin_view {
             $rubricsnew = array($options => $rubricsnew);
 
             // Merge the arrays, prioritising instructor owned arrays.
-            $rubrics = $rubricsnew + $turnitinclass->sharedrubrics;
+            $rubrics = array_merge($rubricsnew, $turnitinclass->sharedrubrics);
         }
 
         $options = array(0 => get_string('no'), 1 => get_string('yes'));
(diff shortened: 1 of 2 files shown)


classes/turnitin_view.class.php
```diff
@@ -98,11 +98,31 @@ public function add_elements_to_settings_form($mform, $course, $location = "acti
...
+
+            $rubrics = array(get_string('attachrubric', 'plagiarism_turnitin') => $instructor->get_instructor_rubrics());
+
+            // Get rubrics that are shared on the account.
+            $turnitinclass = new turnitin_class($course->id);
+            $turnitinclass->sharedrubrics = array();
+            $turnitinclass->read_class_from_tii();
+
+            // This will ensure all rubric keys are integers.
+            $rubricsnew = array(0 => get_string('norubric', 'plagiarism_turnitin'));
```
hperperidis: As discussed you could make this a string key to consistently make this an associative array.
```
rubricsnew = array('0' => get_string('norubric', 'plagiarism_turnitin'));
```

classes/turnitin_view.class.php
```diff
@@ -98,11 +98,31 @@ public function add_elements_to_settings_form($mform, $course, $location = "acti
...
+            foreach ($rubrics AS $options => $rubriclist) {
+                foreach ($rubriclist AS $key => $value) {
+                    $rubricsnew[$key] = $value;
```
hperperidis: then here
```
$rubricsnew[ (string)$key ] = $value;
```

classes/turnitin_view.class.php
hperperidis: just ignore above points. As discussed, this will make no difference.

PR #358, opened 2018-08-20
Fix rubric attach on develop branch.

dwinn:
Fix rubric attach on develop branch.


turnitinplugin_view.class.php
```diff
@@ -117,11 +117,31 @@ public function add_elements_to_settings_form($mform, $course, $location = "acti
...
+
+            $rubrics = array(get_string('attachrubric', 'plagiarism_turnitin') => $instructor->get_instructor_rubrics());
+
+            // Get rubrics that are shared on the account.
+            $turnitinclass = new turnitin_class($course->id);
+            $turnitinclass->sharedrubrics = array();
+            $turnitinclass->read_class_from_tii();
+
+            // This will ensure all rubric keys are integers.
+            $rubricsnew = array(0 => get_string('norubric', 'plagiarism_turnitin'));
```
hperperidis: As discussed this can be turned in a string key to make sure we consistently create an associative array.
i.e.
```
$rubricsnew = array('0' => get_string('norubric', 'plagiarism_turnitin'));

File: turnitinplugin_view.class.php [hunk 2]
```diff
@@ -117,11 +117,31 @@ public function add_elements_to_settings_form($mform, $course, $location = "acti
...
+            foreach ($rubrics AS $options => $rubriclist) {
+                foreach ($rubriclist AS $key => $value) {
+                    $rubricsnew[$key] = $value;
+                }
+            }
+            $rubricsnew = array($options => $rubricsnew);
+
+            // Merge the arrays, prioritising instructor owned arrays.
+            $rubrics = $rubricsnew + $turnitinclass->sharedrubrics;
```
hperperidis: Why not use `array_merge ( array $array1 [, array $... ] )` for this?

turnitinplugin_view.class.php
```diff
@@ -117,11 +117,31 @@ public function add_elements_to_settings_form($mform, $course, $location = "acti
...
+            // Get rubrics that are shared on the account.
+            $turnitinclass = new turnitin_class($course->id);
+            $turnitinclass->sharedrubrics = array();
+            $turnitinclass->read_class_from_tii();
+
+            // This will ensure all rubric keys are integers.
+            $rubricsnew = array(0 => get_string('norubric', 'plagiarism_turnitin'));
+            foreach ($rubrics AS $options => $rubriclist) {
+                foreach ($rubriclist AS $key => $value) {
+                    $rubricsnew[$key] = $value;
```
hperperidis: and here
```
$rubricsnew[(string)$key] = $value;
```

turnitinplugin_view.class.php
dwinn: This doesn't matter as PHP will treat them the same, as discussed.

`$array = array(
    1    => "a",
    "1"  => "b",
    1.5  => "c",
    true => "d",
);
var_dump($array);
array(1) {
  [1]=>
  string(1) "d"
}`

hperperidis: Yes, good point. I agree.

turnitinplugin_view.class.php
hperperidis: so as above, ignore this

PR #360, opened 2018-08-21
Fixing regression for multiple file uploads in split branch.

dwinn:
This pull request fixes a regression in the previous version which prevented multiple file submissions from being queued to Turnitin.

Split branch.


jmcgettrick:
Nice behat test 👍

PR #370, opened 2018-09-18
Fix group submission notice.

dwinn:
Fix group submission notice.


dwinn:
Closing as this is going to require a different fix.

PR #371, opened 2018-09-21
Adding further behat tests.

dwinn:
This pull request adds tests for:

Assignment drafts
Assignment groups
Forums
Workshop
Any file type
Resubmission
EULA
Pseudo user


tests/behat/assignment_groups.feature
```diff
@@ -0,0 +1,107 @@
...
+    And I add "student1 student1" user to "Group 1" group members
+    And I add "student2 student2" user to "Group 1" group members
+    And I am on "Course 1" course homepage
+    And I follow "Test assignment name"
+    And I navigate to "View all submissions" in current page administration
+    And I follow "Test assignment name"
+    Then I should see "Grading summary"
+
+
+   # Note: This test is broken until INT-12029 is completed. Leaving the code here as the test should work once that ticket is fixed.
```
clegg21: Is there a way of letting the tests know that this is expected to fail?

tests/behat/assignment_resubmission.feature
```diff
@@ -0,0 +1,123 @@
...
+    And I log in as "instructor1"
+    And I am on "Course 1" course homepage
+    And I follow "Test assignment name"
+    Then I should see "View all submissions"
+    When I navigate to "View all submissions" in current page administration
+    Then "student1 student1" row "File submissions" column of "generaltable" table should contain "Turnitin ID:"
+    And I log out
+    # Trigger cron as admin for report
+    And I log in as "admin"
+    And I wait "20" seconds
```
clegg21: Will it always be 20 seconds?

tests/behat/workshop.feature
```diff
@@ -0,0 +1,107 @@
...
+      | plagiarism_show_student_report    | 1             |
+    And I am on "Course 1" course homepage
+    And I edit assessment form in workshop "Test workshop" as:"
+      | id_description__idx_0_editor | Aspect1 |
+      | id_description__idx_1_editor | Aspect2 |
+      | id_description__idx_2_editor |         |
+
+  @javascript
+  Scenario: A submission can be queued and sent to Turnitin for a workshop
+    And I change phase in workshop "Test workshop" to "Submission phase"
```
clegg21: I think you're missing a Given here

tests/behat/assignment_groups.feature
dwinn: I don't think so, or at least I can't find anything that allows you to do that. I could just replace all the tags with @expectedfail then the test won't run at all. Or remove it completely and save the code for when we fix the related bug.

tests/behat/behat_plagiarism_turnitin.php
```diff
@@ -130,4 +132,57 @@ public function i_should_see_textcount_instances_of_text_on_the_page($textcount,
...
+     * @param string $coursename
+     * @throws ElementNotFoundException
+     * @throws Exception
+     */
+    public function i_obtain_an_originality_report_for_student_on_modtype_assignmentname_on_course_coursename($student, $modtype, $modname, $coursename)
+    {
+        $reportFound = false;
+        $count = 1;
+        while (!$reportFound) {
+            $this->execute('behat_general::i_run_the_scheduled_task', "\plagiarism_turnitin\\task\update_reports");
```
jmcgettrick: `i_run_the_scheduled_task` I'll need to remember that one rather than navigating to it.

PR #380, opened 2018-10-19
Fix saving of selected rubric.

tlock:
This reverts:

 - 'Remove rubric population from settings page'

This code fails because moodleform can't find posted value in the generated form data without JS.


jmcgettrick:
Closing this due to conflicts.

PR #381, opened 2018-10-22
Reuse user or unenrol if using unique user.

dwinn:
This pull request makes improvements to the way users are handled within behat tests.

- The tests will first attempt to use an existing test account, otherwise create a new one.
- For the EULA test which requires a unique user the test will unenrol the user from Turnitin at the end of the test.

### DIFF
--- a/tests/behat/assignment.feature
+++ b/tests/behat/assignment.feature
@@ -8,8 +8,10 @@ Feature: Plagiarism plugin works with a Moodle Assignment
     Given the following "courses" exist:
       | fullname | shortname | category | groupmode |
       | Course 1 | C1        | 0        | 0         |
-    And I create a unique user with username "student1"
-    And I create a unique user with username "instructor1"
+    And the following users will be created if they do not already exist:
+      | username    | firstname   | lastname    | email                                   |
+      | instructor1 | instructor1 | instructor1 | instructor1_tiibehattesting@example.com |
+      | student1    | student1    | student1    | student1_tiibehattesting@example.com    |
     And the following "course enrolments" exist:
       | user        | course | role    |
       | student1    | C1     | student |
--- a/tests/behat/assignment.feature
+++ b/tests/behat/assignment.feature
@@ -10,8 +10,8 @@ Feature: Plagiarism plugin works with a Moodle Assignment
       | Course 1 | C1        | 0        | 0         |
     And the following users will be created if they do not already exist:
       | username    | firstname   | lastname    | email                                   |
-      | instructor1 | instructor1 | instructor1 | instructor1_tiibehattesting@example.com |
-      | student1    | student1    | student1    | student1_tiibehattesting@example.com    |
+      | instructor1 | instructor1 | instructor1 | instructor1_$account_tiibehattesting@example.com |
+      | student1    | student1    | student1    | student1_$account_tiibehattesting@example.com    |
     And the following "course enrolments" exist:
       | user        | course | role    |
       | student1    | C1     | student |
--- a/tests/behat/behat_plagiarism_turnitin.php
+++ b/tests/behat/behat_plagiarism_turnitin.php
@@ -22,11 +22,14 @@
  */
 
 // NOTE: no MOODLE_INTERNAL test here, this file may be required by behat before including /config.php.
-
 require_once(__DIR__ . '/../../../../lib/behat/behat_base.php');
+require_once(__DIR__ . '/../../vendor/autoload.php');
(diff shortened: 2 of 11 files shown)


tests/behat/behat_plagiarism_turnitin.php
```diff
@@ -214,4 +217,105 @@ public function i_accept_the_turnitin_eula_from_the_ev_if_necessary() {
...
+        }
+        $tablenode = new TableNode($newdata);
+        $this->execute('behat_data_generators::the_following_exist', array('users', $tablenode));
+    }
+
+    /**
+     * @Given /^I unenroll the user account "(?P<student>(?:[^"]|\\")*)" with the role "(?P<role>(?:[^"]|\\")*)" from the class in Turnitin$/
+     * @throws Exception
+     */
+    public function i_unenroll_the_user_account_with_the_role_from_the_class_in_turnitin($student, $role) {
```
jmcgettrick: Nice! 👍

tests/behat/assignment.feature
```diff
@@ -8,8 +8,10 @@ Feature: Plagiarism plugin works with a Moodle Assignment
     Given the following "courses" exist:
       | fullname | shortname | category | groupmode |
       | Course 1 | C1        | 0        | 0         |
-    And I create a unique user with username "student1"
-    And I create a unique user with username "instructor1"
+    And the following users will be created if they do not already exist:
+      | username    | firstname   | lastname    | email                                   |
+      | instructor1 | instructor1 | instructor1 | instructor1_tiibehattesting@example.com |
```
jmcgettrick: Minor point, just something to think about  but could we do the email as a random string instead?

tests/behat/behat_plagiarism_turnitin.php
```diff
@@ -214,4 +217,105 @@ public function i_accept_the_turnitin_eula_from_the_ev_if_necessary() {
...
+     * @Given /^the following users will be created if they do not already exist:$/
+     * @param TableNode $data
+     * @throws Exception
+     */
+    public function the_following_users_will_be_created_if_they_do_not_already_exist(TableNode $data) {
+        $newdata = array();
+        $rowNum = 0;
+        foreach ($data->getRows() as $row) {
+            if (!$rowNum == 0) { // not header row
+                $row[3] = $row[0] . "_". getenv('TII_ACCOUNT') . ".tiibehattesting@example.com";
```
hperperidis: Here it seems you are not reading the data you have defined in the node tables when calling the method. 
You could define email addresses of the form `instructor1_account_tiibehattesting@example.com`.
Then here you could do:
```
foreach($data->getRows() as $row) {
    if (!rowNum == 0) {
        $row[3] = str_replace("_account_", getenv('TII_ACCOUNT'), $row[3]);
    }
}
```
Or something along these lines.

tests/behat/behat_plagiarism_turnitin.php
```diff
@@ -214,4 +217,105 @@ public function i_accept_the_turnitin_eula_from_the_ev_if_necessary() {
...
+        $findmembership = $response->getMembership();
+        $membershipids = $findmembership->getMembershipIds();
+
+        try {
+            foreach ($membershipids as $membershipid) {
+                $membership->setMembershipId($membershipid);
+                $turnitincall->deleteMembership($membership);
+            }
+        } catch (Exception $e) {
+            // ignore exception.
```
hperperidis: We could do some logging here at least to say which memberships failed to be deleted. If you think that would be useful.

tests/behat/assignment.feature
dwinn: We can't use a random string as we end up with the problem of creating a unique user every time (which we want to avoid with this change). Behind this command the account ID is inserted to make the email unique for each account - going to make that more obvious as per Harry's comment.

tests/behat/behat_plagiarism_turnitin.php
dwinn: Yeah sounds good to me.

tests/behat/behat_plagiarism_turnitin.php
dwinn: Contrary to what we spoke about last night - thinking about it more I actually think this would appear in the plugin's API log since deleting the membership is an API request, and that would be sufficient for debugging.

tests/behat/assignment.feature
jmcgettrick: Fair enough! 👍

PR #383, opened 2018-10-23
Conditionally use stdclass for for send messages.

dwinn:
Remove deprecated stdclass for split for send messages.

This is required in preparation for 3.6.

### DIFF
--- a/classes/digitalreceipt/pp_receipt_message.php
+++ b/classes/digitalreceipt/pp_receipt_message.php
@@ -92,6 +92,7 @@ class pp_receipt_message {
      * @return string
      */
     public function build_instructor_message($input) {
+        $message = new stdClass();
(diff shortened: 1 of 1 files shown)


classes/digitalreceipt/pp_receipt_message.php
```diff
@@ -114,9 +115,16 @@ public function build_instructor_message($input) {
      * @return void
      */
     public function send_instructor_message($instructors, $message) {
+        global $CFG;
         $subject = get_string('receipt_instructor_copy_subject', 'plagiarism_turnitin');
 
-        $eventdata = new stdClass();
+        // Pre 2.9 does not have \core\message\message()
```
jmcgettrick: It's harmless having it in but if we aren't supporting 2.9 is it necessary? I'm approving as it's not doing any harm!

PR #443, opened 2019-05-16
Refactored cron method to eliminate the cron method

CharlotteSpinks:


jmcgettrick:
Looks good. Verified.

Issue #463, opened 2019-07-26
Clicking similarity score triggers N Ajax requests


garthwilliamson:
When loading the grading page, an event handler is added N times, where N is the number of similarity scores that are displayed to each similarity score. This results in many extraneous calls to Ajax.php being made, resulting in extreme slowdowns.

jmcgettrick:
Thanks, this has been merged!

PR #464, opened 2019-07-26
Clicking similarity score triggers N Ajax requests

shalugarg-catalyst:
When loading the grading page, an event handler is added N times, where N is the number of similarity scores that are displayed to each similarity score. This results in many extraneous calls to Ajax.php being made, resulting in extreme slowdowns.


garthwilliamson:
fixes #463

danmarsden:
Thanks @garthwilliamson and @shalugarg-catalyst - we noticed some major issues with the ajax.php page on one of our sites this week too and were looking into these as well.

@jmcgettrick - this does seem to slam the server quite badly - would be a good idea to get this fixed asap.

jmcgettrick:
Sorry, had to revert this. It broke something on local environments, will need to investigate further

jonof:
Without knowing what broke, perhaps this would be reliable?
```
    public function load_page_components() {
        global $CFG, $PAGE;

        if (!$PAGE->requires->should_create_one_time_item_now('plagiarism_turnitin')) {
            return;
        }

        $PAGE->requires->js_call_amd('plagiarism_turnitin/open_viewer', 'origreport_open');
        ...
```
At least then you'll get the one invocation until the page is reset and the requires set is flushed.

PR #466, opened 2019-08-22
fix opening the peermark/rubric modals under IE11

jonof:
This enables the Peermark and rubric modal dialogues to open for those unfortunate/obstinate people still using IE11, which doesn't implement the URLSearchParams object.


jmcgettrick:
Thanks @jonof

jmcgettrick:
Closing as IE 11 isn't supported from Moodle 3.10 onwards

PR #495, opened 2020-01-02
Int 13614 change namespace for config

CharlotteSpinks:


classes/forms/turnitin_setupform.class.php
```diff
@@ -227,9 +227,9 @@ public function save($data) {
             if (plugin_supports('mod', $mod, FEATURE_PLAGIARISM)) {
                 $property = "turnitin_use_mod_" . $mod;
                 ${ "turnitin_use_mod_" . "$mod" } = (!empty($data->$property)) ? $data->$property : 0;
-                set_config('turnitin_use_mod_'.$mod, ${ "turnitin_use_mod_" . "$mod" }, 'plagiarism');
+                set_config('turnitin_use_mod_'.$mod, ${ "turnitin_use_mod_" . "$mod" }, 'plagiarism_turnitin');
                 if (${ "turnitin_use_mod_" . "$mod" }) {
-                    set_config('turnitin_use', 1, 'plagiarism');
+                    set_config('turnitin_use', 1, 'plagiarism_turnitin');
```
jmcgettrick: This needs to stay as it was.

db/upgrade.php
```diff
@@ -63,7 +63,7 @@ function xmldb_plagiarism_turnitin_upgrade($oldversion) {
 
         upgrade_dm_successful_uploads();
 
-        upgrade_plugin_savepoint(true, 2013081202, 'plagiarism', 'turnitin');
+        upgrade_plugin_savepoint(true, 2013081202, 'plagiarism_turnitin', 'turnitin');
```
dwinn: I have a feeling these don't need to be changed. Parameters 3 and 4 for this method say:

 * @param string $type The type of the plugin.
 * @param string $plugin The name of the plugin.

Type would be the type of plugin, IE plagiarism, tool, block etc.

So we should change these back, and make sure we test upgrading from an older version.

db/upgrade.php
```diff
@@ -399,7 +399,26 @@ function xmldb_plagiarism_turnitin_upgrade($oldversion) {
...
+        // Update plagiarism to plagiarism_turnitin for consistency.
+        $data = get_config('plagiarism');
+
+        $properties = array("accountid", "apiurl", "secretkey", "enablediagnostic", "enablepeermark",
+             "enableperformancelogs", "repositoryoption", "transmatch", "useerater", "usegrademark", "agreement", "enablepseudo", "pseudofirstname",
+            "pseudolastname", "lastnamegen", "pseudosalt", "pseudoemaildomain", "useanon");
+
+        foreach ($properties as $property) {
+            $key = "plagiarism_turnitin_".$property;
+            if (isset($data->$key)) {
```
dwinn: Should this not be checking for the original key? Since you're checking an object that comes from the database and they haven't been updated yet. So it would be:

```$data->"plagiarism_".$property```

File: db/upgrade.php [hunk 4]
```diff
@@ -402,6 +402,25 @@ function xmldb_plagiarism_turnitin_upgrade($oldversion) {
         upgrade_plugin_savepoint(true, 2019050201, 'plagiarism', 'turnitin');
     }
 
+    if ($oldversion < 2019121709) {
+        // Update plagiarism to plagiarism_turnitin for consistency.
+        $data = get_config('plagiarism');
+
+        $properties = array("accountid", "apiurl", "secretkey", "enablediagnostic", "enablepeermark",
```
dwinn: Just had a thought... this list doesn't contain every setting so we'll need to add the rest to it.

CharlotteSpinks: Added in use_mod_assign, use_mod_forum, use_mod_workshop, just had to make sure these were for PP and not TCA :)

db/upgrade.php
dwinn: You'll have to change the way `$key` is done because in the db `use_mod_assign` as well as forum and workshop are prefixed with just turnitin rather than plagiarism_turnitin. It may be better to just put the full key name within the array and get rid of the concatenation `$key` variable. Can then replace where `$key` is used with `$property`.

db/upgrade.php
```diff
@@ -402,6 +402,25 @@ function xmldb_plagiarism_turnitin_upgrade($oldversion) {
...
+    if ($oldversion < 2019121709) {
+        // Update plagiarism to plagiarism_turnitin for consistency.
+        $data = get_config('plagiarism');
+
+        $properties = array("use_mod_assign", "use_mod_forum", "use_mod_workshop","accountid", "apiurl", "secretkey", "enablediagnostic", "enablepeermark",
+             "enableperformancelogs", "repositoryoption", "transmatch", "useerater", "usegrademark", "agreement", "enablepseudo", "pseudofirstname",
+            "pseudolastname", "lastnamegen", "pseudosalt", "pseudoemaildomain", "useanon");
+
+        foreach ($properties as $property) {
+            $key = "plagiarism_turnitin_".$property;
```
dwinn: You'll have to change the way $key is done because in the db use_mod_assign as well as forum and workshop are prefixed with just turnitin rather than plagiarism_turnitin. It may be better to just put the full key name within the array and get rid of the concatenation $key variable. Can then replace where $key is used with $property.

PR #500, opened 2020-02-13
Remove unneeded ids, add tabindexs and missing config label and chang…

jmcgettrick:
…e colour contrasts


lang/en/plagiarism_turnitin.php
```diff
@@ -36,6 +36,7 @@
 $string['excludequoted'] = 'Exclude Quoted Material';
 $string['excludequoted_help'] = 'This setting allows the instructor to choose to exclude text appearing in the quotes from being checked for matches when generating Similarity Reports. This setting can be overridden in individual Similarity Reports.';
 $string['excludevalue'] = 'Exclude Small Matches';
+$string['excludesmallmatchesvalue'] = 'Exclude Small Matches';
```
dwinn: Any reason we need the same string twice?

jmcgettrick: 🤦‍♂ on myself

PR #511, opened 2020-05-19
change logic to update user role and fix issue reported in git

CharlotteSpinks:


lib.php
```diff
@@ -3045,15 +3045,22 @@ function plagiarism_turnitin_send_queued_submissions() {
...
+        $submission->setRole('Learner');
+
+        if ($queueditem->userid != $queueditem->submitter) {
+
             $instructor = new turnitin_user($queueditem->submitter, 'Instructor');
-            $instructor->edit_tii_user();
 
-            $submission->setSubmitterUserId($instructor->tiiuserid);
-            $submission->setRole('Instructor');
+            // This should be true but in case of an edge case where a user has been deleted in Tii.
```
jmcgettrick, line 3055: "These"

classes/turnitin_user.class.php
```diff
@@ -299,10 +300,12 @@ public function edit_tii_user() {
             try {
                 $turnitincall->updateUser($user);
             } catch (Exception $e) {
-                $toscreen = ($this->workflowcontext == "cron") ? false : true;
+                $toscreen = ($this->workflowcontext == "cron") == false;
```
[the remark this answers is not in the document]
CharlotteSpinks: Done.

lib.php
CharlotteSpinks: Done.

lib.php
```diff
@@ -3045,15 +3045,22 @@ function plagiarism_turnitin_send_queued_submissions() {
...
+
+        if ($queueditem->userid != $queueditem->submitter) {
+
             $instructor = new turnitin_user($queueditem->submitter, 'Instructor');
-            $instructor->edit_tii_user();
 
-            $submission->setSubmitterUserId($instructor->tiiuserid);
-            $submission->setRole('Instructor');
+            // These should be true but in case of an edge case where a user has been deleted in Tii.
+            $updatesuccess = $instructor->edit_tii_user();
```
dwinn, line 3056: These two variables are only used once so I'm not sure we need them. Could just do:

`if ($instructor->edit_tii_user() && $instructor->join_user_to_class($coursedata->turnitin_cid))`

Or if the variables must stay then let's give them better names, such as $isupdated and $isenrolled. It's good practise to prefix boolean variables with `is` or `has`. See https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig

CharlotteSpinks: Yeah this makes sense, I've gone with the first option. Thanks for the variable naming link :D

PR #513, opened 2020-05-29
Use TFS to mark question attempts

jmcgettrick:


classes/modules/turnitin_quiz.class.php
```diff
@@ -61,4 +63,47 @@ public function get_current_gradequery($userid, $moduleid, $itemid = 0) {
...
+    }
+
+    // Set a new mark for a question attempt based on the grade given for the answer in TFS.
+    public function update_mark($attemptid, $identifier, $userid, $grade, $quizgrade) {
+        global $DB;
+
+        $transaction = $DB->start_delegated_transaction();
+
+        $attempt = quiz_attempt::create($attemptid);
+        $quba = question_engine::load_questions_usage_by_activity($attemptid);
```
dwinn, line 79: What is $quba? Might be better to give a more descriptive name. :)

jmcgettrick: `questions_usage_by_activity` - it's called that all through the quiz code. It helped me understand it easier by leaving it as that.

dwinn: Fair point. I wonder why Moodle called it that.

jmcgettrick:
Rebased

PR #520, opened 2020-06-30
Int 12857 delay with large classes

CharlotteSpinks:
Lang string is a placeholder to be replaced by when received from TPG


lib.php
```diff
@@ -192,16 +192,15 @@ public function get_file_upload_errors($offset = 0, $limit = 0, $count = false)
...
-     * This plugin doesn't use it as it would impact negatively on the page loading.
+     * This will alert the user to refresh the assignment when there has been a change in scores.
      *
      * @param $course - Course the module is part of
      * @param $cm - Course module
      * @return string
      */
     public function update_status($course, $cm) {
-        return '';
+        return html_writer::div(get_string('turnitin_score_refresh_alert', 'plagiarism_turnitin'), 'turnitin_score_refresh_alert', array('id' => 'turnitin_score_refresh_alert'));
```
jmcgettrick, line 202: Being picky, that line is probably too long. 

```return a(
                 b,
                 c
            );```

CharlotteSpinks: It is.

CharlotteSpinks: Done.

PR #523, opened 2020-07-06
refactor method as modulename may not always exist

CharlotteSpinks:


lib.php
```diff
@@ -2837,12 +2837,17 @@ function plagiarism_turnitin_coursemodule_standard_elements($formwrapper, $mform
...
+    $modulename = "";
+    if (!empty(get_class($formwrapper))) {
+        $formwrapperclass = explode("_", get_class($formwrapper));
+        $modulename = $formwrapperclass[1];
+    }
+
     $pluginturnitin->get_form_elements_module(
         $mform,
         $context,
-        'mod_'.$formwrapper->get_current()->modulename,
```
dwinn, line 2843: What happens if we replace this with:

`isset($formwrapper->get_current()->modulename) ? 'mod_'.$formwrapper->get_current()->modulename : 'mod_',`

Might not need the new code above.

Also is passing just `mod_` a good idea when there's no module type or would null be better? Depends how Moodle handle the field.

CharlotteSpinks: Yep, that works. Looks much cleaner too. 

we need mod_ to see the plugin settings.

CharlotteSpinks: Changed.

PR #528, opened 2020-07-20
Int 14711 plugin settings appear twice

CharlotteSpinks:


lib.php
```diff
@@ -327,6 +328,7 @@ public function get_form_elements_module($mform, $context, $modulename = "", $so
                 $course = $this->get_course_data($cmid, $COURSE->id);
                 $turnitinview->add_elements_to_settings_form($mform, $course, "activity", $modulename, $cmid, $plagiarismvalues["plagiarism_rubric"]);
             }
+            $settingsdisplayed = true;
```
dwinn, line 331: I think a branch check for <39 may have been a better solution, for consistency with how it was being done in 3.9. But I'm happy with this solution as well if @jmcgettrick is happy.

PR #542, opened 2020-09-14
Fix db inconsistencies

jmcgettrick:
Hi everyone - I've closed all the related PRs and Issues and want to keep everything unified under this one request.

### DIFF
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -444,6 +444,75 @@ function xmldb_plagiarism_turnitin_upgrade($oldversion) {
         set_config('turnitin_use', null, 'plagiarism');
     }
 
+    // This block is to solve a number of inconsistencies between the install and upgrade scripts
+    if ($oldversion < 2020091401) {
+
+        // Set itemid to default to null
+        $table = new xmldb_table('plagiarism_turnitin_files');
+        $field = new xmldb_field('itemid', XMLDB_TYPE_INTEGER, '10', XMLDB_UNSIGNED, false, false, null, 'externalid');
+        if ($dbman->field_exists($table, $field)) {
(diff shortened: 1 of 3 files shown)


jmcgettrick:
@yao9394 @golenkovm @aspark21 could you take a look at this and comment on anything that you spot please?

CharlotteSpinks (approved):

golenkovm:
Hi,
Great. Thanks!
Cheers

jmcgettrick:
Unfortunately I've had to remove the modifications to userid and courseid as those fields can't be updated without removing the indexes.

db/upgrade.php
```diff
@@ -444,6 +444,62 @@ function xmldb_plagiarism_turnitin_upgrade($oldversion) {
...
+            $dbman->change_field_notnull($table, $field);
+        }
+
+        // Set ownerid to not allow null
+        $field = new xmldb_field('ownerid', XMLDB_TYPE_INTEGER, '10', null, XMLDB_NOTNULL, false, null, 'courseid');
+        if ($dbman->field_exists($table, $field)) {
+            $dbman->change_field_notnull($table, $field);
+        }
+
+        // Set turnitin_ctl to not allow null
```
dwinn, line 488: If we set fields to not allow null, what happens if a database does already have null values for these fields? Is this something we need to consider? I think the update to the field might fail if null values are found.

I don't see why courseid, ownerid or turnitin_cid would ever be null but if there's a chance they could be then we need to consider it in this change.

jmcgettrick: Yeah I see your point. They should never be null but they may, I've changed them accordingly!

dwinn (approved):

PR #544, opened 2020-09-18
add setting into setup form and config, add en strings

CharlotteSpinks:


classes/forms/turnitin_setupform.class.php
```diff
@@ -93,11 +93,18 @@ public function definition() {
             1 => get_string('diagnosticoptions_1', 'plagiarism_turnitin'),
             2 => get_string('diagnosticoptions_2', 'plagiarism_turnitin')
         );
+        $turnitinenableremotelogging = array(
+            0 => get_string('remotelogoptions_0', 'plagiarism_turnitin'),
+            1 => get_string('remotelogoptions_1', 'plagiarism_turnitin')
+        );
```
jmcgettrick, line 99: I wouldn't bother with this, just use `$ynoptions` from line 90 for consistency with everything else

CharlotteSpinks: 👍 Will do

lang/en/plagiarism_turnitin.php
```diff
@@ -179,6 +181,8 @@
 $string['diagnosticoptions_0'] = 'Off';
 $string['diagnosticoptions_1'] = 'Standard';
 $string['diagnosticoptions_2'] = 'Debugging';
+$string['remotelogoptions_0'] = 'Off';
+$string['remotelogoptions_1'] = 'On';
```
jmcgettrick, line 185: My other comment means you can get rid of these as well!

jmcgettrick (approved):

dwinn (approved):

Issue #537, opened 2020-08-14
Error on "View the Rubric used for marking" as a student


sleitao:
When submitting an assignment as a student, if I click "View the Rubric used for marking" I get the following error:

Exception - Argument 2 passed to has_capability() must be an instance of context, null given, called in [dirroot]/plagiarism/turnitin/ajax.php on line 162

$plugin->version = 2020072201;

jonof:
This is a fix we're presently evaluating:
```diff --git a/amd/src/rubric.js b/amd/src/rubric.js
index 863d466..fef68fd 100644
--- a/amd/src/rubric.js
+++ b/amd/src/rubric.js
@@ -37,11 +37,13 @@ define(['jquery',
             },
             rubricCreateModal: function(modalType) {
                 var courseid = ($('input[name="course"]')) ? $('input[name="course"]').val() : 0;
+                var cmid = $('input[name="coursemodule"], input[name="id"]').val() || 0;
 
                 ModalFactory.create({
                     type: modalType,
                     templateContext: {
                         courseid: courseid,
+                        cmid: cmid,
                         wwwroot: M.cfg.wwwroot
                     },
                     large: true
@@ -53,4 +55,4 @@ define(['jquery',
                     });
             }
         };
-    });
\ No newline at end of file
+    });

jmcgettrick:
https://github.com/turnitin/moodle-plagiarism_turnitin/pull/550

jonof:
It seems my fix is not enough to solve the error you still get if you click the rubric icon once a submission is no longer editable and there's no form from which the plugin can steal the cmid or courseid. I'll contemplate a more robust solution.

jonof:
This alternative works reliably from our testing: jonof@385f22ec7ec0eade8a2bc13fe6f85cf79d630577

jmcgettrick:
Hey @jonof - can you make a PR for that please?

PR #554, opened 2020-10-09
fix errors when displaying student rubric modal in any circumstance

jonof:
Avoid relying on spying into hidden form fields to find course and/or course module ids, and instead declare data attributes on the element where the event handler is attached.

Refs #537.


jonof:
These were the scenarios I checked:
 - Teacher: rubric manager from the Assignment settings screen, both for unsaved new and existing instances.
 - Student: view rubric before submission.
 - Student: view rubric after submission, both for editable and non-editable submissions.

jmcgettrick (approved):

PR #555, opened 2020-10-21
add method to update and save rubric on ev closing

CharlotteSpinks:


lib.php
```diff
@@ -1932,6 +1932,49 @@ public function sync_tii_assignment($cm, $coursetiiid, $workflowcontext = "site"
         }
     }
 
+    public function update_rubric_from_tii($cm) {
```
dwinn, line 1935: Can we add a PHPdoc? Self explanatory what the method does but Moodle guidelines say methods should be documented.

dwinn (approved):
Looks good!

lib.php
CharlotteSpinks: Done that :)

jmcgettrick (approved):

jmcgettrick:
Verified.

PR #594, opened 2021-07-02
Add filter support to submitter&instructor message

izendegi:
We've detected that the messages send by Turnitin to submitters and instructors don't apply the Moodle filters (we use filter_multilang2, but it may happen also with order filters), this change fixes that.


dwinn (approved):

dwinn:
I'm going to change this to use format_text instead as format_string strips HTML. They both use the filters.

PR #598, opened 2021-09-09
Don't overwrite Moodle grades if there is no grade in TFS

jmcgettrick:
Also, handle test connection after install when no config has been set


dwinn (approved):

mhughes2k:
We seemed to see this happening when the document viewer was launched by a teacher: the grade would be set to null in Moodle, but would the grade also get nullified if it was viewed by a student user?

andrewrowatt-masseyuni:
@mhughes2k we had cases of students causing the grade to get nullified.

Issue #604, opened 2021-10-25
Feedback studio overwrites custom grade scale in Moodle


sharpchi:
In our institution we don't use 100/100 grade scale, but we do use Feedback Studio for feedback. We tell the instructors to not enter a grade in Feedback Studio, but they do sometimes, and this overwrites what they have already entered in Moodle.
Our grade scale is a letter scale A1,A2,A3,A4,B1,B2 etc, but they have the values 1-18 in Moodle's grade selector.

![image](https://user-images.githubusercontent.com/3774280/138720189-ece4437d-13b2-4a20-97d5-9436480bdec7.png)

These are three test scenarios where the grades are overwritten incorrectly:

**Grade in GradeMark** 

1. Grade assignment B1 
2. Open GradeMark 
3. Grade assignment 70 
4. Close GradeMark 
5. Wait for browser to refresh, or manual refresh 
6. Grade should read B1 (Grade is deleted)

**Remove grade in GradeMark** 

1. Ensure assignment has grade 70 in Grademark 
2. Close GradeMark 
3. Wait for browser to refresh, or manual refresh 
4. Enter grade B1 in Moodle 
5. Open GradeMark 
6. Remove grade (i.e. --) 
7. Close GradeMark 
8. Wait for browser to refresh, or manual refresh 
9. Grade should remain B1 (Grade is deleted in version < 2021091501)

**Grade assignment in GradeMark with a score of 10** 

1. Grade assignment in Moodle B1
2. Open GradeMark
3. Grade assignment 10
4. Close GradeMark 
5. Wait for Browser to refresh, or manual refresh 
6. Grade should read B1 (it will be C2) 

So, the plugin makes no adjustment for the grade-scale being used in Moodle, and this is our problem.

I'm going to work on our own solution, which is probably going to allow the Moodle admin to prevent the grade from TII being updated in Moodle (and allow the Instructor to overrule). So I'll probably do a pull request.

PR #605, opened 2021-10-26
ISSUE_604: Optional GradeMark grade import

sharpchi:
A possible work-around for the issue raised in #604 

Allows the site admin or Instructor to disable importing grades from GradeMark. Grades are still stored in the plagiarism_turnitin_files table. It overcomes the issue of incompatible grade scales.

### DIFF
--- a/classes/turnitin_view.class.php
+++ b/classes/turnitin_view.class.php
@@ -215,6 +215,11 @@ class turnitin_view {
             $mform->addElement('select', 'use_turnitin', get_string("useturnitin", "plagiarism_turnitin"), $options);
             $this->lock($mform, $location, $locks);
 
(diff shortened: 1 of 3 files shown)


aspark21:
Yes please, between #598 and #604 it's clear this overwriting of Moodle Grades is a very bad idea and we need a way to switch this off at site level and optionally at activity level

mwillis11:
Sad to see that this PR was not merged in, and the ticket closed - Turnitin overriding our grades is causing immense issues in our platform, especially as we have other processes that kick off when grades are changed, and we use grade scales too. I've had to implement this fix manually in our codebase since 2021. Can we please get this in the plugin?

Issue #482, opened 2019-11-18
DB schema alignment issue: mismatches between install and upgrade scripts


kristian-94:
After installing an older version of the plugin and upgrading, the DB schema will be updated according to the upgrade.php scripts. These scripts leave the DB in a different alignment than a fresh install would.
The output I get from php admin/cli/check_database_schema.php is:
```
$ php admin/cli/check_database_schema.php
-------------------------------------------------------------------------------
plagiarism_turnitin_files
 * column 'itemid' has default 'NULL', expected '0' (I)
 * column 'student_read' has default 'NULL', expected '0' (I)
-------------------------------------------------------------------------------
plagiarism_turnitin_config
 * column 'config_hash' should be NOT NULL (C)
 * column 'config_hash' length is 100, expected 255 (C)
-------------------------------------------------------------------------------
plagiarism_turnitin_users
 * column 'userid' should be NOT NULL (I)
 * column 'turnitin_uid' should be NOT NULL (I)
 * column 'turnitin_utp' should be NOT NULL (I)
-------------------------------------------------------------------------------
plagiarism_turnitin_courses
 * column 'courseid' should be NOT NULL (I)
 * column 'ownerid' should be NOT NULL (I)
 * column 'turnitin_ctl' should be NOT NULL (X)
 * column 'turnitin_cid' should be NOT NULL (I)
-------------------------------------------------------------------------------

```
It's clear that these issues are code bugs, below I've gone through each section and linked the relevant part of the upgrade scripts that are incorrect. It could also be fixed in the install.xml file instead. We'll need to create a new upgrade.php section that goes through and retroactively fixes the schema.

> -------------------------------------------------------------------------------
> plagiarism_turnitin_files
>  * column 'itemid' has default 'NULL', expected '0' (I)

We can see the mismatch between the install and upgrade scripts:
https://github.com/turnitin/moodle-plagiarism_turnitin/blob/master/db/install.xml#L15
https://github.com/turnitin/moodle-plagiarism_turnitin/blob/master/db/upgrade.php#L238

 

> * column 'student_read' has default 'NULL', expected '0' (I)

We can see the mismatch between the install and upgrade scripts:
 https://github.com/turnitin/moodle-plagiarism_turnitin/blob/master/db/install.xml#L26
 https://github.com/turnitin/moodle-plagiarism_turnitin/blob/master/db/upgrade.php#L259
 
 

> -------------------------------------------------------------------------------
[...]

PR #585, opened 2021-05-24
Set userid and courseid columns to be NOT NULL (master branch)

golenkovm:
This is a fix for #517

PR #611, opened 2021-12-16
Fix issue#482 plagiarism_turnitin_config

TomoTsuyuki:
Fix the issue #482


TomoTsuyuki:
This needs to be deployed with https://github.com/turnitin/moodle-plagiarism_turnitin/pull/585 to fix all of issues for #482

dwinn:
Hi @TomoTsuyuki ,

Thank you for your contribution.

I am a little concerned that this does not check for existing values in the database before changing the index. See how #585 does so. Maybe that needs to be in the code here to prevent issues if there is a null value for that field.

TomoTsuyuki:
Hi @dwinn ,

Thanks for checking my PR.
I updated the code to check the field before changing index as #585 .

dwinn:
Hi @TomoTsuyuki,

Thank you for the quick update, was just about to make the change myself as we're keen to get this out in our next release but then I saw you updated. We are in the final stages of preparing a release with some other community pull requests, currently aiming for early next week.

dwinn (approved):

TomoTsuyuki:
Hi @dwinn 

Thank for merging.
Hope the update will save many people :)

Issue #499, opened 2020-01-31
Users not prompted to accept EULA when using Forum inline editing feature


rayhinton:
Hello Turnitin,

**Observed behavior:** if a user writes a post with the new Moodle Forum inline editing feature, then they are never shown the TII EULA during the writing process.

**Expectation:** new users should be prompted to accept the Turnitin End User License Agreement when they are making their first forum post.

**Consequences:** it is more likely that a user can write multiple forum posts before realizing they need to accept the TII EULA, which means that their initial posts will not be processed and receive similarity scores.

**Proposed solution:** incorporate EULA prompt into the inline editor.

**Details:**

Moodle 3.7.3+ (Build: 20200103)
Turnitin plagiarisim plugin version 3.1+, 2019121701
Boost theme (and custom Boost-based themes)

Moodle version 3.7 introduced the Forum inline editing feature (I've also been calling it "Quick post" or "Quick reply"). Please see:

https://docs.moodle.org/37/en/Using_Forum#Replying_to_discussions
https://docs.moodle.org/37/en/New_features

When new users are starting their first action in an activity where the Turnitin Plagiarism plugin is enabled, they are usually prompted to accept the TII EULA. However, with the Forum inline editor, the experience is different. The EULA prompt is not shown to users if they are using the inline editor.

Of course, it is still shown to people in the usual way if they click the "Advanced" link to view all the posting options. Further, they can also accept the EULA by going to one of their completed posts, and finding the prompt there.

So, there is nothing preventing the user from accepting the EULA *eventually*. However, the path to accepting the prompt feels substantially different with the inline editor. Before, users were shown the EULA prompt very explicitly, before they could even start typing anything in their first forum post. Now, they need to seek out the EULA prompt.

Because of this, we have received a significant increase in users who are experiencing confusion with accepting the EULA. We are doing our best to guide them through it, and indeed give them instructions before they encounter it.

This is a problem because users are more likely to make multiple posts before realizing that they have not accepted the TII EULA. Since there is not a convenient way to resubmit posts for Turnitin processing when this happens, this means that a user could end up having several posts that never get processed by Turnitin.

Can the EULA prompt be incorporated into the inline editor? Conceptually, that seems like the most straightforward solution. I attached a couple examples I mocked up (apologies for my poor editing skills).

Of course, I imagine that this kind of issue could possibly require changes in the Moodle Forum module. I would be happy to start a Moodle Tracker issue, if that is the case.

[...]

PR #615, opened 2022-03-04
Show the 'accept EULA' prompt on new replies made in-page to a forum

jonof:
Currently the "Your file has not been submitted to Turnitin. Please click here to accept our EULA." message only shows on Forum posts when the page is first loaded, because that's when _eula.js_ takes notice and shows `.pp_turnitin_eula` elements. If a user performs an in-page reply, any newly-introduced `.pp_turnitin_eula` element within the newly-added post remains hidden until the page is next reloaded fully. Since Moodle 3.8, [an event is triggered by the Forum plugin](https://github.com/moodle/moodle/blob/MOODLE_38_STABLE/mod/forum/amd/src/inpage_reply.js#L169) when a new post has been added, and this patch listens for that event to then show the `.pp_turnitin_eula` element that may have come into being if the user hasn't previously accepted the EULA.


jonof:
Note: ought to solve issue #499

dwinn (approved):

PR #623, opened 2022-06-16
Quiz attempt grades may be incorrectly set

jonof:
This bug can occur in two observed ways:

1. Worst case: by causing incorrect grades to be calculated for a student's quiz attempt, hopefully nonsensical enough to be obvious.
2. Best case: by doing nothing while logging an error in the Turnitin activity log:

    `2022-06-15 16:00:36 +1000 (API_ERROR) - There was an error when trying to get a submission from Turnitin`
    `File: /.../question/engine/datalib.php | Line: 478 | Message: Coding error detected, it must be fixed by a programmer: Failed to load questions_usage_by_activity 3 | Code: 0`

The first situation is disturbing because opening a similiarity report and closing it can be enough to corrupt students' calculated total grades without it being particularly obvious to the teacher what has happened. A Quiz regrade is then needed to bring totals back into order. This happens because `quiz_attempts` and `question_usages` table IDs likely overlap such that a given `quiz_attempt` ID is also a valid `question_usages` ID.

The second situation means that at worst Grademark-given grades are not applied the student's response.

### DIFF
--- a/classes/modules/turnitin_quiz.class.php
+++ b/classes/modules/turnitin_quiz.class.php
@@ -76,7 +76,7 @@ class turnitin_quiz {
         $transaction = $DB->start_delegated_transaction();
 
         $attempt = quiz_attempt::create($attemptid);
-        $quba = question_engine::load_questions_usage_by_activity($attemptid);
+        $quba = question_engine::load_questions_usage_by_activity($attempt->get_uniqueid());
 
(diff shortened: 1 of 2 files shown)


dwinn (approved):

dwinn:
Hi @jonof ,

Thank you for your contribution. We will aim to get this released shortly.

PR #624, opened 2022-06-20
Only load cmid when its not already present

aolley:
Without this, get_links calls load up the cmid of quizzes every time its called
(which can be hundreds of times) - ignoring when its already been given the
cmid.

This resolves a performance issue when viewing quiz results where a lot of
users have taken a quiz with essay questions.


mhughes2k:
this is an observation that has been levelled at the Turnitin Plugin before (on the assignment module), so anything that reduces the amount of work that the plugin is doing during the get_links() call (which should be basically restricted to just reading from the database in IMHO) would definitely be supported by me! (this is for the benefit of the TII devs!)

danmarsden:
The turnitin get_links code is why I wrote this plugin:

https://github.com/catalyst/moodle-local_disableplagiarism

Would definitely be nice to see some attention to some performance improvements.

dwinn (approved):

PR #632, opened 2022-09-13
Rubric manager size

dwinn:
This resizes the iframe used for displaying rubrics. This is the best way to do this without interfering with the CSS Moodle uses for modals. Grunt was complaining about using the package tag in the JS files so I've removed that as it's unnecessary.


ixiongtii:
lgtm

styles.css
```diff
@@ -367,6 +367,11 @@ iframe html{background:none;}
     height: 680px;
 }
 
+.iframe-ltilaunch-rubric {
+    width: 840px;
```
MartinLenord, line 371: would it be worth making this 100% then have an min-width of 840px?

dwinn: I've changed it to this. Visibly I can't see any difference between this and what I had it as though.

MartinLenord (approved):

PR #645, opened 2023-01-09
Add translated matching score checks

ixiongtii:


sahmad-tii (approved):
LGTM

vendor/Integrations/phpsdk-package/src/SubmissionSoap.php
```diff
@@ -46,6 +46,7 @@ public function readSubmission($submission)
                 $tiiSubmission->setAuthorUserId($soap->resultRecord->result->personSourcedId);
                 $tiiSubmission->setDate($soap->resultRecord->result->date);
                 $tiiSubmission->setOverallSimilarity($soap->resultRecord->result->resultScore->textString);
+                $tiiSubmission->setTranslatedOverallSimilarity($soap->resultRecord->result->extension->extensionField[6]->fieldValue);
```
baholladay, line 49: I'm not familiar with the extension fields... do you need to add any checks for whether `extension` or `extensionField[6]` exists before setting this field?

ixiongtii:
One thing that's interesting is that all the `fieldValue` are strings. So if a submission has the`TranslatedOverallSimilarity` field name, but doesn't have an actual score to go with that, then the results give an empty string instead of `null`. And php will read this empty string as 0 when type-casted to an int (which it does in `lib.php`)

PR #647, opened 2023-02-23
fix/create_submission_for_duplicate_text

ixiongtii:


lib.php
```diff
@@ -2582,7 +2582,8 @@ public function event_handler($eventdata) {
                 $eventdata['other']['content'] = $qa->get_response_summary();
 
                 // Queue text content.
-                $identifier = sha1($eventdata['other']['content']);
+                // adding slot to sha hash to create unique assignments for duplicate text based on it's id
+                $identifier = sha1($eventdata['other']['content'].$slot);
```
baholladay, line 2586: is `slot` never empty?

ixiongtii: `$slot` should never be empty, because it refers to each question in a quiz. The only case if `$slot` is empty, would be if a quiz doesn't have a question. And if there are no questions, then a Tii assignment would not be created and queued.

PR #650, opened 2023-03-21
fix/resubmit_submission_eula

ixiongtii:
Addresses the issue where a user declines the EULA and submits a submission, then later accepts the EULA and is not able to resubmit the same paper/submission.

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -2307,8 +2307,8 @@ class plagiarism_plugin_turnitin extends plagiarism_plugin {
         $coursedata = $this->get_course_data($cm->id, $cm->course);
         $user = new turnitin_user($author, "Learner");
         $user->join_user_to_class($coursedata->turnitin_cid);
-        $eulaaccepted = ($user->useragreementaccepted == 0) ? $user->get_accepted_user_agreement() : $user->useragreementaccepted;
-        if ($eulaaccepted != 1) {
+        $eula_accepted = ($user->useragreementaccepted == 0) ? $user->get_accepted_user_agreement() : $user->useragreementaccepted;
+        if ($eula_accepted != 1) {
             return true;
         }
         // Check if file has been submitted before.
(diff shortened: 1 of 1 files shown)


lib.php
```diff
@@ -2303,6 +2303,14 @@ public function queue_submission_to_turnitin($cm, $author, $submitter, $identifi
         $attempt = 0;
         $tiisubmissionid = null;
 
+        // If the EULA hasn't been accepted, don't save submission and don't submit to Tii
+        $coursedata = $this->get_course_data($cm->id, $cm->course);
+        $user = new turnitin_user($author, "Learner");
+        $user->join_user_to_class($coursedata->turnitin_cid);
+        $eulaaccepted = ($user->useragreementaccepted == 0) ? $user->get_accepted_user_agreement() : $user->useragreementaccepted;
+        if ($eulaaccepted != 1) {
+            return true;
```
baholladay, line 2312: is there some indicator for the users (instructor?) that show this submission wasn't queued b/c of no EULA?

ixiongtii: Yep. There is still a warning that is shown if the user hasn't accepted the EULA. Also, when a student doesn't accept the EULA, the submission is still submitted to Moodle, but not queued to Tii

lib.php
```diff
@@ -2303,6 +2303,14 @@ public function queue_submission_to_turnitin($cm, $author, $submitter, $identifi
         $attempt = 0;
         $tiisubmissionid = null;
 
+        // If the EULA hasn't been accepted, don't save submission and don't submit to Tii
+        $coursedata = $this->get_course_data($cm->id, $cm->course);
+        $user = new turnitin_user($author, "Learner");
+        $user->join_user_to_class($coursedata->turnitin_cid);
+        $eulaaccepted = ($user->useragreementaccepted == 0) ? $user->get_accepted_user_agreement() : $user->useragreementaccepted;
```
sahmad-tii, line 2310: Looks correct to me, but I would suggest renaming the variable $eulaaccepted to $eula_accepted to follow the recommended naming convention for PHP variables. 
        Also in other areas in this file I can snake_case being followed.

sahmad-tii (approved):

PR #665, opened 2023-07-06
Submission removed observer

tuanngocnguyen:
This  adds an observer to handle submission removal event, follow up the suggestion (number 2) by Tomo on

https://github.com/turnitin/moodle-plagiarism_turnitin/issues/556

### DIFF
--- a/version.php
+++ b/version.php
@@ -19,7 +19,7 @@
  * @copyright 2012 iParadigms LLC
  */
 
-$plugin->version = 2023033001;
+$plugin->version = 2023033002;
 
 $plugin->release = "3.5+";
 $plugin->requires = 2018051700;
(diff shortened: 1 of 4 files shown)


jack-tii:
@tuanngocnguyen Thanks for your contribution! Could you rebase against the develop branch?

version.php
```diff
@@ -19,7 +19,7 @@
  * @copyright 2012 iParadigms LLC
  */
 
-$plugin->version = 2023033001;
+$plugin->version = 2023033002;
```
jack-tii, line 22: We update the version number based on our the date of release, we would need to revert this change befor this could be merged

tuanngocnguyen:
Hi @jack-tii ,

Thanks for your note,

I have open a new PR with a different branch instead of re-basing the current branch (it may be being used)

Hence I am closing this PR

The new PR is https://github.com/turnitin/moodle-plagiarism_turnitin/pull/730

PR #661, opened 2023-04-14
Fixes #642: Remove minified js from amd/src

petersistrom:
Resolves #642

Issue #642, opened 2022-10-18
Fix Minified Javascript code


TheBryanOpen228:
When updating the plugin to the new 2022092101 Version for moodle 4.0, a Moodle core PhpUnit test was failing, see the issue log here: 
```
core_requirejs_testcase::test_requirejs
Failed asserting that '/vagrant/www/moodle/plagiarism/turnitin/amd/src/modal_eula_launch.min.js' does not contain ".min".
 /vagrant/www/moodle/lib/tests/requirejs_test.php:76
 /vagrant/www/moodle/lib/phpunit/classes/advanced_testcase.php:91
```
This is because in a commit, [(Turnitin commit)](https://github.com/turnitin/moodle-plagiarism_turnitin/commit/2f7f756d9118585ccb9823e417eda564a7a27e4f) the minification of the javascript files was performed erroneously, creating minified files in the src source folder, and minified from the minified in the build folder inside the plagiarism_turnitin plugin, this inconvenience produced a failure in that PhpUnit test of moodle core, since there are not supposed to be any minified files inside the src folder.

jack-tii:
Thanks for the catch! This was resolved by #667

PR #667, opened 2023-07-31
remove extra min js files

ixiongtii:


danmarsden:
note this is a duplicate of #661 and is also reported in #642

ixiongtii:
Yep, we'll be merging this to develop first. Thanks peteristrom for the pr

PR #553, opened 2020-10-05
Issue #448 Move depracted mod methods into hooks.

andrewmadden:
get_form_elements_module > plagiarism_turnitin_coursemodule_standard_elements
save_form_elements > plagiarism_turnitin_coursemodule_edit_post_actions

PR #682, opened 2023-12-08
Fix: Use proper Modal Dialogue for eula modal

ixiongtii:


kkouchtii (approved):
wow, you are good, Javascript + PHP

andrii-ilin:
@ixiongtii I think we need to add a `build` folder to `gitignore` file, builds should not be a part of the repo...

ixiongtii:
> @ixiongtii I think we need to add a `build` folder to `gitignore` file, builds should not be a part of the repo...

Ok, I've added `amd/build` to the gitignore. Can I get another review? @andrii-ilin

danmarsden:
Please don't do that! Moodle plugins do include the build folder. (I know this differs from some other projects) in moodles case the built amd files should be included and removing them from git will cause all sorts of problems for end-users as they will not expect it.

andrii-ilin:
@danmarsden unexpected :-) , thanks for letting us know, @ixiongtii sorry, my bad, I didn't expect that build folders and files should be commited

andrii-ilin (approved):

danmarsden:
thanks @andrii-ilin - glad I jumped in before you merged that in - although I'm surprised to see work in this space going on while other important issues like #553 are being neglected which is needed asap so that sites using Moodle 4.3 can use Turnitin (at the moment sites that have upgraded to Moodle 4.3 cannot use Turnitin code without adding that patch manually themselves.)

PR #709, opened 2024-05-30
Fix/quickmarks peermarks launch

andrii-ilin:
* fix path to modules


ixiongtii (approved):
Rename js references. lgtm

PR #711, opened 2024-06-03
Fix/user_agreement_launch

andrii-ilin:
* hide redundand user agreement button


ixiongtii (approved):
were you able to test this and the eula still launches appropriately where needed?

andrii-ilin:
> were you able to test this and the eula still launches appropriately where needed?

yes, tested with new and existed users, with some different scenarious

PR #724, opened 2024-07-25
Remove asterisk in quiz reference in README.md.

lucaboesch:
This reference is not needed any more.


jack-tii (approved):
Thanks, this was in need of an update

PR #728, opened 2024-07-26
Moodle Plugin CI, and Moodle coding style

lucaboesch:
In these commits, I wanted to prove that the plagiarism_turnitin plugin could well be rewritten/reformatted to adhere to Moodle coding guidelines.
The submitted did not completely fix PHPUnit and Behat tests, though.

Thanks for taking into consideration. If desired, I can also divide into smaller pieces, of course.


jack-tii:
Thanks very much for this PR, we should definitely be following the Moodle style guides.
We are currently having internal discussions about how best to add Github workflows to our plugins as described in https://moodlehq.github.io/moodle-plugin-ci/
Merging this would therefore have to wait until after our next release, which is currently planned for August 1st. I will keep you updated regarding our decisions around this, as this is a feature I would love to have.

PR #729, opened 2024-07-27
Adding LICENSE file resolves #719.

### LINKED ISSUE #719: Please could you add a LICENCE file
All the code is GPL3 - the comment at the top of each source file says so.

However, there is no top-level LICENCE file. Adding one would be good practice. Thanks for considering this.

lucaboesch:


jack-tii (approved):
LGTM, thanks!

PR #730, opened 2024-07-29
Submission removed observer

tuanngocnguyen:
This adds an observer to handle submission removal event, follow up the suggestion (number 2) by Tomo on

https://github.com/turnitin/moodle-plagiarism_turnitin/issues/556

Hi @jack-tii ,

Thanks for your note on:

https://github.com/turnitin/moodle-plagiarism_turnitin/pull/665

I have created another PR with a new branch in case (submission_removed_observer is being used)
The new branch has already rebased against 'develop'

### DIFF
--- a/classes/observer.php
+++ b/classes/observer.php
@@ -132,6 +132,20 @@ class plagiarism_turnitin_observer {
     }
 
     /**
+     * Handle the assignment submission_removed event.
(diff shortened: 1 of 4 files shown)


jack-tii (changes requested):
Could you revert the change to version.php? We only change this when we release a new version

PR #731, opened 2024-07-29
Moodle coding style fixes (without Github actions)

lucaboesch:
Referring to https://github.com/turnitin/moodle-plagiarism_turnitin/pull/728 and the discussion there.


lucaboesch:
I've updated the coding style issues again, and it would be great if they could be merged.

classes/privacy/provider.php
```diff
@@ -25,15 +25,16 @@
 
 namespace plagiarism_turnitin\privacy;
 
-defined('MOODLE_INTERNAL') || die();
-
```
jack-tii, line 29: I trust that you know what you're doing by removing this, but for my own information and out of curiosity what's the reasoning for adding/removing these in a couple of places? Why for example does the privacy provider need to be accessible from scripts external to Moodle?

jack-tii (changes requested):
Thanks so much for your continued support, this was a pretty big undertaking! There are 2 minor things I've commented on, if we can get some clarification and one minor change I'm happy to merge this now that we're at the start of our next release cycle

classes/privacy/provider.php
lucaboesch: Thanks @jack-tii !
I did remove those lines.
In regards to the 'MOODLE_INTERNAL' this might be of value: https://www.youtube.com/watch?v=TaVwRjxN_3E
Tools local/moodlecheck and local/codechecker which I heavily use do also point out omitted and superfluous usages of the command.
Best,
Luca

jack-tii:
Closing as #728 was merged

Issue #712, opened 2024-06-04
Failing unit tests in Moodle 4.3


saidisalem:
plagiarism_turnitin is failing to pass unit testing with Moodle 4.3. There are some changes to quiz classes in Moodle core, please see the trace below:

```
plagiarism_turnitin_quiz_testcase::test_update_mark
Unexpected debugging() call detected.
Debugging: Class 'quiz' has been renamed for the autoloader and is now deprecated. Please use 'mod_quiz\quiz_settings' instead.
* line 153 of /lib/classes/component.php: call to debugging()
* line 58 of /plagiarism/turnitin/tests/modules/turnitin_quiz_test.php: call to core_component::classloader()
* line 1548 of /vendor/phpunit/phpunit/src/Framework/TestCase.php: call to plagiarism_turnitin_quiz_testcase->test_update_mark()
* line 1154 of /vendor/phpunit/phpunit/src/Framework/TestCase.php: call to PHPUnit\Framework\TestCase->runTest()
* line 81 of /lib/phpunit/classes/advanced_testcase.php: call to PHPUnit\Framework\TestCase->runBare()
* line 728 of /vendor/phpunit/phpunit/src/Framework/TestResult.php: call to advanced_testcase->runBare()
* line 904 of /vendor/phpunit/phpunit/src/Framework/TestCase.php: call to PHPUnit\Framework\TestResult->run()
* line 675 of /vendor/phpunit/phpunit/src/Framework/TestSuite.php: call to PHPUnit\Framework\TestCase->run()
* line 675 of /vendor/phpunit/phpunit/src/Framework/TestSuite.php: call to PHPUnit\Framework\TestSuite->run()
* line 675 of /vendor/phpunit/phpunit/src/Framework/TestSuite.php: call to PHPUnit\Framework\TestSuite->run()
* line 653 of /vendor/phpunit/phpunit/src/TextUI/TestRunner.php: call to PHPUnit\Framework\TestSuite->run()
* line 144 of /vendor/phpunit/phpunit/src/TextUI/Command.php: call to PHPUnit\TextUI\TestRunner->run()
* line 97 of /vendor/phpunit/phpunit/src/TextUI/Command.php: call to PHPUnit\TextUI\Command->run()
* line 98 of /vendor/phpunit/phpunit/phpunit: call to PHPUnit\TextUI\Command::main()
* line 122 of /vendor/bin/phpunit: call to include()
[...]

PR #734, opened 2024-08-05
Fix phpunit with Moodle 4.3 and PHP 8.2.

timhunt:
We were seeing PHPunit failures after installing this plugin into Moodle 4.3 (with PHP 8.2).

This is one way to fix it. May not be the most elegant, but I hope it is OK.

Among other things, this fixes issue #712.

I tested on Moodle 4.1 with PHP 7.4 to verify that I had not broken anything.

I hope this is helpful @carl-hostrander.

### DIFF
--- a/classes/modules/turnitin_quiz.class.php
+++ b/classes/modules/turnitin_quiz.class.php
@@ -74,8 +74,13 @@ class turnitin_quiz {
         global $DB;
 
         $transaction = $DB->start_delegated_transaction();
+        if (class_exists('\mod_quiz\quiz_attempt')) {
+            $quizattemptclass = '\mod_quiz\quiz_attempt';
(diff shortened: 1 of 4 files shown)


jack-tii (approved):
Thanks! I'm currently trying to improve our CI processes for these plugins so that we can find things like this faster ourselves... Hopefully won't have to be relying on others for long.

timhunt:
Thanks!

And, in case you have not seen this bit of Moodle development news: Right now there is a know issue with https://moodlehq.github.io/moodle-plugin-ci/. See https://github.com/moodlehq/moodle-plugin-ci/issues/309 and https://github.com/nvm-sh/nvm/issues/3405 for details. Hopefully that gets sorted soon.

PR #735, opened 2024-08-13
lib: skip fake files immediately

jrchamp:
If the file being checked is the special character `.`, then we should `continue;` before trying to read the file or even open a file handle.

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -2644,6 +2644,8 @@ class plagiarism_plugin_turnitin extends plagiarism_plugin {
                     plagiarism_turnitin_activitylog('File not found: '.$pathnamehash, 'PP_NO_FILE');
                     $result = true;
                     continue;
+                } else if ($file->get_filename() === '.') {
+                    continue;
                 } else {
                     try {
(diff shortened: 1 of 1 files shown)


lib.php
```diff
@@ -2644,6 +2644,8 @@ public function event_handler($eventdata) {
                     plagiarism_turnitin_activitylog('File not found: '.$pathnamehash, 'PP_NO_FILE');
                     $result = true;
                     continue;
+                } else if ($file->get_filename() === '.') {
```
usqfowlerj, line 2647: FWIW `$file->is_directory()` does the same test.

jack-tii (approved):

lib.php
jrchamp: Excellent suggestion! Plus, if Moodle were to ever change things, your recommendation would continue to work, which is better.

Reference: https://github.com/moodle/moodle/blob/a75365f2e470fb9a9ef58f3045905acdddc4e410/lib/filestorage/stored_file.php#L363-L365

PR #737, opened 2024-08-14
Merge 660 into develop

jack-tii:
This PR was accidentally merged into master, so merging to develop here

### DIFF
--- a/classes/modules/turnitin_quiz.class.php
+++ b/classes/modules/turnitin_quiz.class.php
@@ -87,7 +87,8 @@ class turnitin_quiz {
         foreach ($attempt->get_slots() as $slot) {
             $answer = $attempt->get_question_attempt($slot)->get_response_summary();
             // Check if this is the slot the mark is for by matching content.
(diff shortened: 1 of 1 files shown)


classes/modules/turnitin_quiz.class.php
```diff
@@ -87,7 +87,8 @@ public function update_mark($attemptid, $identifier, $userid, $grade, $quizgrade
         foreach ($attempt->get_slots() as $slot) {
             $answer = $attempt->get_question_attempt($slot)->get_response_summary();
             // Check if this is the slot the mark is for by matching content.
-            if (sha1($answer.$slot) == $identifier) {
+            $answerslot = $answer ? $answer.$slot : $answer;
```
jrchamp, line 90: If $answer, when coerced to a boolean, is false, it seems like we should use `$slot` on its own.

```suggestion
            $answerslot = $answer ? $answer.$slot : $slot;
```

dclusker-tii (approved):

PR #741, opened 2024-09-05
Replace deprecated download_as_dataformat

jack-tii:
As per https://tracker.moodle.org/browse/MDL-68500, download_as_dataformat is deprecated since Moodle 3.9 and we should be using \core\dataformat::download_data instead


jack-tii:
Superseded by #752

Issue #515, opened 2020-06-10
Performance issues with scheduled tasks


danmarsden:
The scheuled tasks appear to be extremely in-efficient.

In particular the "update_reports" task which is set to run every 5 min by default.

for one of our sites - it doesn't usually do less than 5000 db reads and 800 writes every time, which means that's it's doing 720,000 reads and 115,200 writes over a 24hr period.

We had a much bigger situation yesterday (guessing a bunch of assignments were submitted) where it was averaging around 25,000 reads, 7000 writes every 5min - so it hit almost 4 Million db reads and 1 million writes over the same 24 hr period.

Looking at this code here:
https://github.com/turnitin/moodle-plagiarism_turnitin/blob/74f979a9fa7efd655d26e7e5749dde3a8fab7ce4/lib.php#L1921

I can see several places where db queries are called inside loops when you should really be able to get a lot more information in the initial db query.

thanks!

agrowe:
After looking into performance issues with our scheduled tasks in cron we are also seeing ~7000 DB queries each time `plagiarism_turnitin\task\update_reports` is run.

Not as bad as customcert which was almost doing 27k queries every minutes, but still needs to be handled better.

danmarsden:
it's fun when it runs "on the hour" too... because that just happens to be when people like to schedule exams/timed quizzes - right when you don't really want the db server doing heaps of other work...

agrowe:
Here's the last 24 hours of DB queries for our Moodle's sorted by query count.

This is the top 7 queries. 2/3/5/7 all look similar on the graph so are most likely caused by this plugin (but some are used in Moodle elsewhere), or exacerbated by this plugin.

Every 5 minutes, 6681 minimum queries logged in the cron output for this task.

We only return 124 rows for the `get_records_select` too, so something funny going on with this task.

![Untitled](https://user-images.githubusercontent.com/1534918/98179380-bb97d000-1f52-11eb-9e13-f77fcca7248c.png)

rlorenzo:
Seeing the same here. These are our last 3 runs:

42498 reads
9061 writes

42554 reads
9073 writes

39580 reads
8413 writes

I am changing our task runtime to be every 10 minutes instead of every 5 minutes. That function does run a lot of things in loops... See https://github.com/turnitin/moodle-plagiarism_turnitin/blob/master/lib.php#L1954

danmarsden:
[...]

PR #743, opened 2024-09-06
Optimising DB calls in cron

jack-tii:
Addresses #515

Currently we are doing a whole load of unnecessary DB reads when running the cron job to send queued files to tii.

This PR does 2 things:

1. Fetch the coursemodule instance and module name upfront instead of in a loop. This eliminates the need to call get_coursemodule_from_id for each submission, removing 4 DB reads per submission.
2. Cache module settings up front - we only need one DB read per module for this, not one per submission. Therefore we can fetch them all in advance and read them from an array later.

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -2041,7 +2041,12 @@ class plagiarism_plugin_turnitin extends plagiarism_plugin {
           ['success', 1]
         );
 
-        echo('$submissions: ' . json_encode($submissions, JSON_PRETTY_PRINT) . "\n");
+        $modulesettings = [];
+        foreach ($submissions as $tiisubmission) {
+          if (!array_key_exists($tiisubmission->cm, $modulesettings)) {
+            $modulesettings[$tiisubmission->cm] = $this->get_settings($tiisubmission->cm);
+          }
+        }
 
         // Add submission ids to the request.
         foreach ($submissions as $tiisubmission) {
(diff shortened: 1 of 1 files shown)


dclusker-tii (approved):
lgtm

lib.php
```diff
@@ -2015,63 +2015,62 @@ public function cron_update_scores() {
...
+          'SELECT PTF.*, 
+          CM.instance AS instance,
+          M.name AS modname
+          FROM {plagiarism_turnitin_files} PTF
+          JOIN {course_modules} CM ON CM.id = PTF.cm
+          JOIN {modules} M ON M.id = CM.module
+          WHERE statuscode = ?
+          AND ( similarityscore IS NULL OR duedate_report_refresh = 1 )
+          AND ( orcapable = ? OR orcapable IS NULL )
+          ORDER BY externalid DESC',
```
baholladay, line 2029: run this query by the DB team and confirm that this will work for mysql, postgres, Microsoft SQL Server, and  oracle. Also confirm that a generic "JOIN" is preferable over "RIGHT JOIN"

lib.php
```diff
@@ -2015,63 +2015,62 @@ public function cron_update_scores() {
...
+            $now = strtotime('now');
+            $dtdue = (!empty($moduledata->duedate)) ? $moduledata->duedate : 0;
+            if ($tiisubmission->duedate_report_refresh != 1 && $now >= $dtdue && $now < strtotime('+1 day', $dtdue)) {
+                $this->set_duedate_report_refresh($tiisubmission->id, 1);
+            }
 
-                if (!isset($reportsexpected[$cm->id])) {
-                    $plagiarismsettings = $this->get_settings($cm->id);
-                    $reportsexpected[$cm->id] = 1;
+            if (!isset($reportsexpected[$tiisubmission->cm])) {
```
baholladay, line 2050: change to:
```
if (!isset($modulesettings[$tiisubmission->cm]) {
    if (!isset($modulesettings[$tiisubmission->cm]['plagiarism_compare_institution'])) {
    
    }
...
}|
```

lib.php
```diff
@@ -2015,63 +2015,62 @@ public function cron_update_scores() {
...
+          AND ( orcapable = ? OR orcapable IS NULL )
+          ORDER BY externalid DESC',
+          ['success', 1]
         );
 
-        // Add submission ids to the request.
+        $modulesettings = [];
         foreach ($submissions as $tiisubmission) {
+          if (!array_key_exists($tiisubmission->cm, $modulesettings)) {
+            $modulesettings[$tiisubmission->cm] = $this->get_settings($tiisubmission->cm);
```
baholladay, line 2036: what does this return if it doesn't exist? It may be better to not set this unless you have a valid response to set

lib.php
```diff
@@ -2015,63 +2015,62 @@ public function cron_update_scores() {
...
-                    );
+            // Only add the submission to the request if we are expecting an originality report.
+            if ($reportsexpected[$tiisubmission->cm] == 1) {
+                $submissionids[] = $tiisubmission->externalid;
 
-                    if (!isset(array_flip($assignmentids)[$moduleturnitinconfig->value])) {
-                        $assignmentids[] = $moduleturnitinconfig->value;
-                    }
+                if (!isset(array_flip($assignmentids)[$modulesettings[$tiisubmission->cm]->turnitin_assignid])) {
+                    $assignmentids[] = $modulesettings[$tiisubmission->cm]->turnitin_assignid;
```
baholladay, line 2073: can you provide some clarity on why we no longer need to call `plagiarism_turnitin_config`?

PR #748, opened 2024-09-10
More database optimisation

jack-tii:
Caching module data and using INNER JOIN

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -2016,13 +2016,17 @@ class plagiarism_plugin_turnitin extends plagiarism_plugin {
         $reportsexpected = array();
         $assignmentids = array();
         
+        // Grab all plagiarism files where all the following conditions are met:
+        // 1. The file has been successfully sent to TII
+        // 2. The submission is ready to recieve a similarity score (either it doesn't already have a similarity score or it's set to regenerate)
+        // 3. The module associated with the submission still exists
         $submissions = $DB->get_records_sql(
           'SELECT PTF.*, 
           CM.instance AS instance,
           M.name AS modname
           FROM {plagiarism_turnitin_files} PTF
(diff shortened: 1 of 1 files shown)


jack-tii:
Tagging this with help wanted as it's a critical part of the code and any changes we make have to be bulletproof.
I'm intending on testing against a variety of databases, but concerned about any edge cases we might encounter - collation issues perhaps. If anyone in the community could offer insights this would be extremely helpful.
This PR is the second half of our changes to this cron, the first half is here: #743

NeillM:
Hi Jack,

I have taken a look at the change and it looks as though the intent was to do less database queries during the cron.

For the module information it looks as though the whole record is being stored for each activity instance while only the duedate is being used, on a large busy site this could potentially be many activities and the records contain blobs which could potentially hold large amounts of data. It is probably worth limiting the fields you get back from the database to just those that will be used by [specifying them in the third field of get_record](https://moodledev.io/docs/4.5/apis/core/dml#get_record) this way it should use much less memory (especially as sometimes images get pasted into activity summaries as base64 encoded data so each one can be huge).

It might also be worth considering if you can get the data you want from the plagiarism_turnitin_config table without using a query in the loop as well, since that seems to undo at least some of the things you did in the other query that will now be called less. Can you JOIN or LEFT JOIN it into the query that fetches the submissions for example? That would probably significantly reduce the total number of queries made.

I hope this is useful.

NeillM:
On the duedate field not being present in all activities it would be possible to check if it existed using [$DB->get_columns()](https://github.com/moodle/moodle/blob/main/lib/dml/moodle_database.php#L1165)

dclusker-tii (approved):

PR #753, opened 2024-09-25
Allow non submitting group members to access DV

jack-tii:
Previously when a group submission was made, only the submitting user would have access to DV. The other users would get a similarity score, but the DV link would not show up.

This PR checks whether an assignment is configured to accept group submissions. If it is, we open the DV link as the submitting user instead of as the currently logged in user.


jack-tii:
Closing, as this is not possible within the TFS permission system.

PR #761, opened 2024-10-23
Fix ajax.php error 404

jack-tii:
When opening the grading page the rubric manager tries to launch because we are unnecessarily including the javascript module that loads the list of available rubrics. This is already loaded separately for the settings page, so we don't need to load it twice. Additionally, the URL in this file was relative, so would sometimes point to `/mod/plagiarism/turnitin/ajax.php`, which does not exist.


jack-tii:
We should be using `M.cfg.wwwroot` for all URLs within javascript modules

PR #767, opened 2024-11-04
Change rubric ID to integer

jack-tii:
We are currently sending a blank string as the rubric ID to TFS instead of the integer 0 when no rubric is attached to an assignment. This fixes a bug where it is impossible to detach rubrics from PP TFS assignments.


jack-tii:
Also added some extra error logging for this error so we can catch further instances in the future

TurnitinPaulina (approved):
From what i know of tfs, that 0 would have definitely caused bugs and this should fix it, looks good to me

Issue #770, opened 2024-11-14
We’ve noticed high latency... or Undefined property $version during install


leonstr:
If you install plagiarism_turnitin 2024072401 on an existing Moodle 4.5 site the message _We’ve noticed high latency in your connection to Turnitin. To help you remedy this, we’ve turned off automatic updates to the inbox. Refresh to update submissions_ appears at the top of the _Current release information_ page (admin/index.php) during plugin install.

If you install Moodle 4.5 with the plagiarism_turnitin 2024072401 already present in the source code directory you get  _Warning: Undefined property: stdClass::$version in /var/www/moodle/plagiarism/turnitin/db/hooks.php on line 25_ at the top of the _Installation_ page (admin/index.php).

This appears to be because the before_standard_top_of_body_html_generation hook is used but is missing a [check for the installation or upgrade states](https://moodledev.io/docs/4.5/apis/core/hooks#registering-of-hook-callbacks):

> Callbacks are executed during system installation and all upgrades, the callback methods must verify the plugin is in correct state. Often the easiest way is to use function during_initial_install() or version string from the plugin configuration.

leonstr:
Fix merged.

PR #771, opened 2024-11-14
770 Add install state check for Hooks API

leonstr:
Proposed fix for #770


jack-tii:
This is a good catch, we missed testing this during initial install
Thanks, I'll merge this

PR #774, opened 2024-11-25
762 Fix: Mismatched anonymous define

leonstr:
Move DataTables to vendorjs which appears to fix "Mismatched anonymous define" RequireJS error in browser console when tool_ally is installed.


jack-tii (changes requested):
Thanks for the PR!
Could you revert the change to $plugin->version? We only update this when doing a release.
Other than that, the change looks good. I'd like to get this in with our upcoming release.

leonstr:
Force pushed with `$plugin->version` reverted.

jack-tii (approved):

PR #775, opened 2024-11-25
773 Fix: jQuery is not a function

leonstr:
Fix error in browser console error which occurs with Moodle 4.5 on plugin settings page (plagiarism/turnitin/settings.php).

This seems to work correctly with Moodle 4.4 too.


jack-tii (approved):
I was about to put up the same fix, you beat me to it by about 5 minutes
Thanks!

PR #777, opened 2024-11-29
Using notifications API for latency warnings

jack-tii:
Right now when we detect high latency in the Turnitin connection we notify the user by adding a chunk of html at the top of the page. This is a clunky approach for a few reasons:

1. We don't have control over where on the page the message ends up getting displayed, it's just placed right at the top where it doesn't look very good
2. This approach relies on custom CSS to hide the error message when not needed, so it's not compatible with custom themes
3. We have to use the hooks API for this on Moodle 4.4+ and standard callbacks on older versions - this introduces unnecessary tech debt as we now have to test separately on 2 different branches and potentially debug new issues introduced by the hooks API

As we are now using AMD modules we can switch over to using the notifications API instead. This greatly simplifies our code and allows using custom themes.


TurnitinPaulina (approved):
Looks like we are now using notification, looks good

Issue #789, opened 2024-12-20
Exception shown when a student submits their quiz attempt


anieminen:
Hi,

We are seeing a problem after upgrading plagiarism_turnitin plugin. When Turnitin is activated in quiz settings and a user tries to submit their quiz attempt, an exception is thrown and the attempt is not submitted. 

![kuva](https://github.com/user-attachments/assets/af6ba04d-4d7e-4aba-8bda-7f586889a226)

Moodle version: 4.1.15
plagiarism_turnitin version: 4.1+ (2024120501)

Debug info:
Error code: generalexceptionmessage
Stack trace:
    line 2617 of /plagiarism/turnitin/lib.php: Error thrown
    line 159 of /plagiarism/turnitin/classes/observer.php: call to plagiarism_plugin_turnitin->event_handler()
    line ? of unknownfile: call to plagiarism_turnitin_observer::quiz_submitted()
    line 155 of /lib/classes/event/manager.php: call to call_user_func()
    line 75 of /lib/classes/event/manager.php: call to core\\event\\manager::process_buffers()
    line 835 of /lib/classes/event/base.php: call to core\\event\\manager::dispatch()
    line 2310 of /mod/quiz/attemptlib.php: call to core\\event\\base->trigger()
    line 2220 of /mod/quiz/attemptlib.php: call to quiz_attempt->fire_state_transition_event()
    line 2030 of /mod/quiz/attemptlib.php: call to quiz_attempt->process_finish()
    line 93 of /mod/quiz/view.php: call to quiz_attempt->handle_if_time_expired()

davidscotson:
We hit this same issue, on 4.1.15 and plagiarism_turnitin version: 4.1+
(2024121901)

From some initial investigation it seems as if the change to add namespaces in this commit might be the cause:

https://github.com/turnitin/moodle-plagiarism_turnitin/commit/d7d61731a340ea333db066e8444a183e8624287f

A slightly earlier change checks for the existence of that class before using it in order to support both 4.2 upwards and older version of Moodle, 4.1 being the current LTS version:

[c00e3438aa620df217805a827fd66e308f8b3be8](https://github.com/turnitin/moodle-plagiarism_turnitin/commit/c00e3438aa620df217805a827fd66e308f8b3be8)

This seems related to the quiz namespaces changes mentioned in the mod_quiz changelog for 4.2:

https://github.com/moodle/moodle/blob/main/mod/quiz/upgrade.txt

[...]

PR #794, opened 2025-01-27
Add version check for mod_quiz namespace

jack-tii:
In Moodle 4.2 the PHP class quiz_attempt was moved out of the global namespace and into mod_quiz

Currently our plugin works for newer Moodle versions (4.2+) but we need to support Moodle 4.1. Therefore we need to add a check for the current Moodle version, and use the namespace version on Moodle 4.2+.

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -2510,7 +2510,7 @@ class plagiarism_plugin_turnitin extends plagiarism_plugin {
      * @return bool result
      */
     public function event_handler($eventdata) {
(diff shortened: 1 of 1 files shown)


jack-tii:
This should fix #789

rrusso (commented):
please review this, it's bad copy pasta and you're missing the mod/quiz/locallib include above the 1st call to the class.

PR #856, opened 2025-07-23
Fix #855 - use xmldb editor to reformat install.xml file.

### LINKED ISSUE #855: invalid install.xml file.
Moodle now includes a unit test that checks the validity of the install.xml file in all plugins (see MDL-85666)

The install.xml file in the code here appears to be hand-rolled instead of using the xmldb editor and so now fails these unit tests.

This should be fixed using the Moodle xmldb editor (under admin > development > XMLDB.)

Please hit the "load" link for your plugin install.xml file in the moodle xmldb editor, and then make Moodle think there is an update to the file (easiest way to do this is to change the description of a table - add a full stop or something similar.) Then get XMLDB to "Save" the updated file which will remove all the old invalid xml from your file.

danmarsden:
I've used xmldb editor to reformat this file - however I had to remove the invalid calls to "ONDELETE="cascade"  in the file before xmldb editor would load the existing file.


jack-tii (approved):
Thanks, I did this for the integrity plugin also: https://github.com/turnitin/moodle-plagiarism_turnitinsim/pull/216

PR #863, opened 2025-08-01
Quiz attempt hash updates

jack-tii:
For online text quiz attempts the hash lookup is failing as it's being calculated differently in `event_handler` vs `get_links`.
Also, hash collisions are possible when users upload the same text to two or more different quiz questions.
This PR adds a new hashing method that fixes both of these problems. We now calculate the hash for an online text quiz question attempt by concatenating the question type, the user id, the cmid, the slot, and the attempt id to create a unique string which is then hashed.
Example of generating a hash using the new method: `sha1(quiz_attempt user21 cm33 slot4 attempt1)`
This PR also adds a bit of extra logging for cron errors and implements looking submissions up by the previously generated hash to maintain backward compatibility.

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -832,7 +832,16 @@ class plagiarism_plugin_turnitin extends plagiarism_plugin {
                 }
                 $content = $moduleobject->set_content($linkarray, $cm);
                 if ($submissiontype === 'quiz_answer') {
-                  $identifier = sha1($linkarray['area'].$linkarray['itemid']);
+
+                  if (class_exists('\mod_quiz\quiz_attempt')) {
+                      $quizattemptclass = '\mod_quiz\quiz_attempt';
+                  } else {
+                      $quizattemptclass = 'quiz_attempt';
+                  }
+                  $attempt = $quizattemptclass::create($linkarray["area"]);
+
+                  $identifier = sha1('quiz_attempt user'.$attempt->get_userid().' cm'.$cm->id.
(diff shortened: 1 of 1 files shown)


lib.php
```diff
@@ -3490,12 +3510,17 @@ function plagiarism_turnitin_send_queued_submissions() {
...
                     $errorcode = 14;
                     break;
                 }
+
+                // Attempt to find the matching slot for the queued item.
+                // For each slot, check whether the hash matches.
                 foreach ($attempt->get_slots() as $slot) {
                     $qa = $attempt->get_question_attempt($slot);
-                    if ($queueditem->identifier == sha1($queueditem->itemid.$slot)) {
+                    if ($queueditem->identifier == sha1('quiz_attempt user'.$attempt->get_userid().' cm'.$cm->id.
```
TurnitinPaulina, line 3522: we now make the identifier be composed of all these values?

jack-tii: Previously the identifier was the sha1 hash of the text content of the submission, with the slot appended on the end. For the test "This is a submission" and the slot "1" You get "This is a submission1"
This is kind of flaky, as the slot is not unique. If another student uploaded the exact same content to another quiz, but the same question slot, the hashes would collide. Another way this can happen is if the student reattempts the quiz.
Therefore I decided to instead create a unique string containing the CM, attempt number, etc. and hash it.

TurnitinPaulina (approved):

PR #865, opened 2025-08-14
Note only for assignment or coursework

TurnitinPaulina:


classes/turnitin_view.class.php
```diff
@@ -380,9 +380,11 @@ public function add_elements_to_settings_form($mform, $course, $location = "acti
                 $mform->addElement('hidden', 'plagiarism_rubric', '');
                 $mform->setType('plagiarism_rubric', PARAM_RAW);
             }
-
-            $mform->addElement('html', html_writer::tag('div', get_string('anonblindmarkingnote', 'plagiarism_turnitin'),
-                                                                                ['class' => 'tii_anonblindmarkingnote']));
+            // Only relevant for Assignments and Coursework
+            if ($modulename == 'mod_assign' || $modulename == 'mod_coursework') {
```
jack-tii, line 384: *Extremely* minor nit - for good practice we should use `===` for comparing string literals, since the `==` operator performs type coercion

jack-tii (approved):
Looks good!

PR #869, opened 2025-08-20
checking max attempts instead of deprecated const

TurnitinPaulina:


classes/modules/turnitin_assign.class.php
```diff
@@ -140,7 +140,7 @@ public function is_resubmission_allowed($assignid, $reportgenspeed, $submissiont
         // If resubmissions are enabled in a Turnitin sense.
         if ($reportgenspeed > 0) {
             // If the attempt reopened method is none, or an attempt has not been reopened/has previous submission.
```
jack-tii, line 142: nit: need to update this comment

classes/modules/turnitin_assign.class.php
jack-tii, line 1: At line 138 I think the `require_once($CFG->dirroot . '/mod/assign/locallib.php');` is no longer needed - we were including it for the ASSIGN_ATTEMPT_REOPEN_METHOD_NONE define

PR #883, opened 2025-09-05
using usage id instead of attempt id

TurnitinPaulina:


jack-tii (approved):
Simple bug fix, approved

PR #923, opened 2025-12-09
Converted submissions queue to using adhoc tasks.

jarad-barker:
- Add adhoc worker tasks which process the queue in batches
  - Add a cleanup task to reset queued submissions after a configurable timeframe

### DIFF
--- a/classes/forms/turnitin_setupform.class.php
+++ b/classes/forms/turnitin_setupform.class.php
@@ -172,6 +172,17 @@ class turnitin_setupform extends moodleform {
         $mform->addElement('static', 'plagiarism_turnitin_agreement_desc', null, get_string('pp_agreement_desc',
             'plagiarism_turnitin'));
 
+        $mform->addElement('text', 'plagiarism_turnitin_cron_submissions_limit', get_string('turnitin_cron_submissions_limit', 'plagiarism_turnitin'));
+        $mform->setType('plagiarism_turnitin_cron_submissions_limit', PARAM_INT);
+        $mform->setDefault('plagiarism_turnitin_cron_submissions_limit', 100);
+
+        $mform->addElement('duration', 'plagiarism_turnitin_cron_submissions_cleanup', get_string('turnitin_cron_submissions_cleanup', 'plagiarism_turnitin'));
+        $mform->addElement('static', 'turnitin_cron_submissions_cleanup_desc', null, get_string('turnitin_cron_submissions_cleanup_desc', 'plagiarism_turnitin'));
+        $mform->setDefault('plagiarism_turnitin_cron_submissions_cleanup', 10800);
+
+        $mform->addElement('textarea', 'plagiarism_turnitin_agreement', get_string("pp_agreement", "plagiarism_turnitin"), 'wrap="virtual" rows="10" cols="50"');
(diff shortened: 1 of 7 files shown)


classes/forms/turnitin_setupform.class.php
```diff
@@ -172,6 +172,17 @@ public function definition() {
...
 
+        $mform->addElement('text', 'plagiarism_turnitin_cron_submissions_limit', get_string('turnitin_cron_submissions_limit', 'plagiarism_turnitin'));
+        $mform->setType('plagiarism_turnitin_cron_submissions_limit', PARAM_INT);
+        $mform->setDefault('plagiarism_turnitin_cron_submissions_limit', 100);
+
+        $mform->addElement('duration', 'plagiarism_turnitin_cron_submissions_cleanup', get_string('turnitin_cron_submissions_cleanup', 'plagiarism_turnitin'));
+        $mform->addElement('static', 'turnitin_cron_submissions_cleanup_desc', null, get_string('turnitin_cron_submissions_cleanup_desc', 'plagiarism_turnitin'));
+        $mform->setDefault('plagiarism_turnitin_cron_submissions_cleanup', 10800);
+
+        $mform->addElement('textarea', 'plagiarism_turnitin_agreement', get_string("pp_agreement", "plagiarism_turnitin"), 'wrap="virtual" rows="10" cols="50"');
```
dmitriim, line 183: @jarad-barker is this a duplication of line 170?

classes/task/cleanup_queued_submissions.php
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+namespace plagiarism_turnitin\task;
+
+defined('MOODLE_INTERNAL') || die();
+
+/**
+ * Cleanup queued submissions to Turnitin.
+ */
+class cleanup_queued_submissions extends \core\task\scheduled_task {
```
dmitriim, line 33: "Clean up" here is a bit confusing as usually clean up is associated with deleting something. Here we clearly not doing that and just reset records.  Would be good to add a bit more information about this task and maybe think about better name?

lang/en/plagiarism_turnitin.php
```diff
@@ -319,3 +320,6 @@
 $string['dbexport'] = 'Database Export';
 $string['dbexporttable'] = 'Export {$a} data';
 $string['invalidtablename'] = 'Table {$a} could not be exported';
+$string['turnitin_cron_submissions_limit'] = 'Turnitin cron submissions limit';
+$string['turnitin_cron_submissions_cleanup'] = 'Turnitin cron submissions cleanup timeframe';
+$string['turnitin_cron_submissions_cleanup_desc'] = 'Cleanup task will reset all queued submissions that have not been sent after this period of time';
```
dmitriim, line 325: This description generates more questions than answers. Would be useful to add more information like what this all actually mean in the context of processing the queue? Will the records be processed again? Excluded from processing? Pushed up to the queue? it's not clear after reading. 
Also as stated above - "clean up" seems to be a wrong wording.

dmitriim:
@jarad-barker spending more time on this PR it feels like some bits are missing. Would you please be able to revisit and resubmit if possible?

PR #932, opened 2026-02-13
Remove redundant eula check

jack-tii:
When deciding whether to queue a submission to Turnitin, we must check whether the submitter has accepted the eula. Currently we do this by making a call to TFS. However this is redundant as we already store the eula acceptance state in the database. This PR replaces the API call with checking the database.


dvicente308 (approved):

jack-tii:
Duplicate

PR #940, opened 2026-05-19
Allow submissions to be sent to TFS with invalid file types

jack-tii:
Fixes an issue observed where if "Allow any file type" is set, an error will still be shown when attempting to submit an unsupported file type.


stemple-tii (approved):
Precog ESRs

PR #942, opened 2026-05-20
Show eula before posting to forum

jack-tii:
This brings the plugin into parity with the integrity plugin by showing the eula prior to first submission to a forum.


dvicente308 (approved):
Looks good

PR #953, opened 2026-06-30
Enable ad-hoc processing for submission queue

jack-tii:
Adds a new ad hoc task, `adhoc_send_submission`. This is an ad hoc task that sends a single submission to Turntin.
The user can choose whether to this enable ad hoc task. If it's disabled, the submissions will be sent sequentially as they were before. If it's enabled, the submissions will be processed in parallel.
Each time the scheduled task runs (every 5 minutes) instead of grabbing 100 submissions and trying to send them sequentially, we grab all queued submissions and spin up an ad hoc task for each one. The site admin is free to configure a limit on how many of these tasks can run at any given time.

---

> [!NOTE]
> **Medium Risk**
> Changes how queued submissions are drained (unlimited fan-out vs 100/run) and relies on Moodle ad-hoc concurrency limits; misconfiguration or duplicate task queuing could increase Turnitin API load or race on the same queue rows.
> 
> **Overview**
> Adds an admin **Enable ad-hoc task processing for the submission queue** setting (default off) and persists `plagiarism_turnitin_enableadhocsubmissions` with other plugin config.
> 
> When enabled, the scheduled **send submissions** task no longer uploads files itself: it loads **all** `queued`/`pending` rows (no per-run cap), verifies Turnitin connectivity, and enqueues one **`adhoc_send_submission`** task per item. That ad-hoc task allows concurrent runs and calls the new shared **`plagiarism_turnitin_send_single_submission`** helper. When disabled, behavior stays the same as before—still capped at **`PLAGIARISM_TURNITIN_CRON_SUBMISSIONS_LIMIT` (100)** and processed sequentially in the scheduled task.
> 
> The large **`lib.php`** change is mostly extracting the per-item upload pipeline into **`plagiarism_turnitin_send_single_submission`**, with loop `continue` semantics replaced by early `return` so both paths share one implementation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 776a3d9cc9b5e9a4e013a7a6f79a191b0b6d1171. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/classes/forms/turnitin_setupform.class.php
+++ b/classes/forms/turnitin_setupform.class.php
@@ -46,6 +46,7 @@ class turnitin_setupform extends moodleform {
         global $DB, $CFG;
 
         $config = plagiarism_plugin_turnitin::plagiarism_turnitin_admin_config();
+        $ynoptions = [0 => get_string('no'), 1 => get_string('yes')];
 
         $mform = $this->_form;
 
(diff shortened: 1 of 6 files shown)


dvicente308 (approved):
Looks good

PR #959, opened 2026-07-15
INT-24996 Fix forum submission identifiers

dvicente308:
> [!NOTE]
> **Medium Risk**
> Changes how forum submissions are deduplicated and looked up in Turnitin; incorrect hashing could miss reports or queue duplicates, though legacy identifiers are still queried for display.
> 
> **Overview**
> Forum Turnitin rows are keyed by **content-only** `sha1` hashes, which can collide across users or activities and can disagree with what gets queued on submit. This change uses a **composite hash** (`forum_post user` + userid + cm id + content) when queuing forum text in `event_handler` and when resolving originality links in `get_links`.
> 
> **`get_links`** also keeps **`oldidentifier = sha1(content)`** so existing `plagiarism_turnitin_files` records still match, following the same new/old pattern used for quiz answers. Plugin **version** is bumped to `2025103111`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5c16bb61d5abc1faebf3a83d7c6a6085063bf7aa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -820,8 +820,10 @@ class plagiarism_plugin_turnitin extends plagiarism_plugin {
                   $identifier = sha1('quiz_attempt user'.$attempt->get_userid().' cm'.$cm->id.
                                      ' slot'.$linkarray["itemid"].' attempt'.$attempt->get_attempt_number());
                   $oldidentifier = sha1($content.$linkarray["itemid"]);
-                }
-                else {
+                } else if ($submissiontype === 'forum_post') {
+                  $identifier = sha1('forum_post user'.$linkarray['userid'].' cm'.$cm->id.' '.$content);
+                  $oldidentifier = sha1($content);
+                } else {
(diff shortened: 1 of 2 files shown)


jack-tii (approved):
Looks good, if we always have the old identifier as a fallback this is a fairly low risk change I think

jack-tii (changes requested):
The same user can submit to a forum multiple times, we'll need to include some other disambiguation than the cm and user ID

jack-tii:
After review I think there's no way to disambiguate further - for other activity module types we get an `itemid` in the linkarray, but no such id is provided for forum online text submissions. This might actually be the best we can do.

jack-tii (approved):

